PR #82, opened 2018-05-22
docs/CNAME: Add openmetrics.io

RichiH:


brian-brazil:
:+1:

RichiH:
Neat! I think I will simply let .xyz lapse, so no worries.

PR #93, opened 2018-07-16
Start working on ID

RichiH:

### DIFF
--- a/draft-hartmann-openmetrics-00
+++ b/draft-hartmann-openmetrics-00
@@ -17,6 +17,13 @@ Given all this, OpenMetrics is an effort to establish a new and universal standa
 Based on the Prometheus exposition format which already saw massive adoption with more than 300 exporters, TODO links TODO more content.
 This extensive support with TODO different implementations of data ingestors by competing companies and projects shows that there's already an established consensus which is merely refined and made offical with the help of this document.
 
+# Metrics and time series
+
+Metrics are a specific kind of system data.
+All system states are expressed as numerical values; counts of things, ENUMs, and boolean states being common examples.
+
+TODO what are metrics, how do they differ from other data
+
 # Metric naming
 
 TODO suffices, etc
@@ -34,31 +41,39 @@ TODO rationale why text is default
 
 Special care has been taken to make the wire formats ease to generate, to ingest, and to understand..
 OpenMetrics is a type 3 grammar, making writing quick and small parsers possible.
-UTF-8 compresses well, and protobuf is already binary.
+Text compresses well, and protobuf is already binary.
 
 ## Text format
 
 Metric names MUST appear as a group.
+If a metricname has appeared in a group before it MUST NOT be repeated after that group finishes.
+
 
 ### ABNF
 
 ABNF as per RFC 5234, UTF8-char as per RFC 3629.
 
-group = *metadata 1*metricdata
+data = *group
 
-metadata = # SP ("HELP" / "TYPE" / TODO) SP metadata-text LF
+group = *metadata *metricdata
 
-metadata-text = UTF8-char *(UTF8-char / SP) # TODO - how to do UTF-8?
+metadata = ( "" / # SP ( "HELP" / "TYPE" / "UNIT" ) SP metadata-text LF )
+
+metadata-text = UTF8-char *(UTF8-char / SP)
 
 metricdata = metricname labelset SP datapoint *1(SP timestamp) *1exemplar LF
 
-metricname = ALPHA *(ALPHA / DIGIT / "_")
+metricname = ( ALPHA / ":" ) *( ALPHA / DIGIT / "_" / ":" )
 
 labelset = ( "" / "{}" / labelset-labels)
 
-labelset-labels = { (1*UTF8-char "=" DQUOTE 1*UTF8-char DQUOTE) *("," 1*UTF8-char "=" DQUOTE 1*UTF8-char DQUOTE) }
+labelname = ALPHA *( ALPHA / DIGIT / "_" )
 
-datapoint = 1*DIGIT *1("." *DIGIT *1( "e" ( "+" / "-" ) 2DIGIT ) )
+labelset-labels = { ( labelname "=" DQUOTE 1*UTF8-char DQUOTE) *("," labelname "=" DQUOTE 1*UTF8-char DQUOTE) }
+
+float = 1*DIGIT *1( "." 1*DIGIT *1( "e" ( "+" / "-" ) 1*3DIGIT ) )
+
+datapoint = ( float / "+Inf" / "-Inf" / "NaN" )# TODO https://github.com/OpenObservability/OpenMetrics/pull/93#discussion_r213579928
 
 timestamp = 1*DIGIT # TODO adapt to timestamp discussion
 
@@ -69,6 +84,7 @@ exemplar = "#" SP 1*(UTF8-char / SP) # put in length limit
 ````
 # HELP go_gc_duration_seconds A summary of the GC invocation durations.
 # TYPE go_gc_duration_seconds summary
+# UNIT seconds
 go_gc_duration_seconds{quantile="0"} 4.9797e-05
 go_gc_duration_seconds{quantile="0.25"} 9.026e-05
 go_gc_duration_seconds{quantile="0.5"} 0.000109604
--- a/draft-hartmann-openmetrics-00
+++ b/draft-hartmann-openmetrics-00
@@ -1,0 +1,64 @@
+# Rationale
+
+The monitoring world is fragmented and there is no up-to-date standard.
+The closest to an universal monitoring standard we have is SNMP.
+While it is reasonably well-supported in the networking world, it failed to gain wide adoption beyond this field.
+SNMP also carries over many engineering trade-offs which made sense at the time, but are not matching well to today's trade-offs; a prime example would be the need to support ASN.1 with its many intricacies.
+
+While there have been other efforts, none of them gained any wide-spread traction.
+Furthermore, most standardization attempts have been foiled by trying to establish themselves in a niche and expanding from there.
+
+More and more fields of IT are being swept up by commoditization, currently aka foo-as-code.
+HTTP endpoints are replacing TCP/IP-based services, an UTF-8 has established itself as the universal non-binary encoding.
+Finally, metrics-based monitoring is overhauling large parts of the monitoring landscape due to their scalability and adaptability.
+
+Given all this, OpenMetrics is an effort to establish a new and universal standard.
+Based on the Prometheus exposition format which already saw massive adoption with more than 300 exporters, TODO links TODO more content.
+This extensive support with TODO different implementations of data ingestors by competing companies and projects shows that there's already an established consensus which is merely refined and made offical with the help of this document.
+
+# Data transmission
+
+There are two wire formats specified in OpenMetrics: The text format and the protobuf format.
+All client implementations MUST be able to emit text format; they MAY also emit protobuf if negotiated.
+All server implementations MUST be able to ingest text format; they MAY also ingest protobuf if negotiated.
+All implementations MUST default to the text format.
+
+TODO rationale why text is default
+
+# Wire formats
+
+Special care has been taken to make the wire formats efficient.
+The three main considerations are CPU time, RAM usage, and network utilization, in this order.
+OpenMetrics is a type 3 grammar, making writing quick and small parsers possible.
+UTF-8 compresses well, and protobuf is already binary.
+
+## Text format
+
+
+TODO more than one data point per time series -> timestamps required
+
+## protobuf format
+
+# Security Considerations
+
+# IANA Considerations
+
+TODO Ports, but probably only one for emitting & receiving
+
+TODO MIME / content types for plain text & proto
+
+
+# general TODOs
+
+TODO push vs pull
+
+TODO versioning
+
+TODO test suite & reference implementations
+
+TODO appendix with supporters
+
+TODO labels / n-dimensional matrices vs trees?
+
+TODO metric names unique with a certain label set
+
--- a/draft-hartmann-openmetrics-00
+++ b/draft-hartmann-openmetrics-00
@@ -9,7 +9,7 @@ While there have been other efforts, none of them gained any wide-spread tractio
 Furthermore, most standardization attempts have been foiled by trying to establish themselves in a niche and expanding from there.
 
 More and more fields of IT are being swept up by commoditization, currently aka foo-as-code.
-HTTP endpoints are replacing TCP/IP-based services, an UTF-8 has established itself as the universal non-binary encoding.
+HTTP endpoints are replacing TCP/IP-based services, and UTF-8 has established itself as the universal non-binary encoding.
 Finally, metrics-based monitoring is overhauling large parts of the monitoring landscape due to their scalability and adaptability.
 
 Given all this, OpenMetrics is an effort to establish a new and universal standard.
--- a/draft-hartmann-openmetrics-00
+++ b/draft-hartmann-openmetrics-00
@@ -66,19 +66,27 @@ metadata = ( "" / # SP ( "HELP" / "TYPE" / "UNIT" ) SP metadata-text LF )
 
 metadata-text = *(UTF8-char / SP)
 
+# TODO https://github.com/OpenObservability/OpenMetrics/pull/93#discussion_r248005922
 metricdata = metricname labelset SP datapoint *1(SP timestamp) *1exemplar LF
 
-#TODO https://github.com/OpenObservability/OpenMetrics/pull/93#discussion_r203009500
-metricname = ( ALPHA / ":" ) *( ALPHA / DIGIT / "_" / ":" )
+metricname-namespace = ( ALPHA / "_" / ":" ) *( ALPHA / DIGIT / "_" / ":" ) ( ALPHA / DIGIT / ":" )
+
+metricname-namespace-suffix = "__"
+
+metricname-suffix = ( ALPHA / ":" ) *( ALPHA / DIGIT / "_" / ":" )
+
+metricname-standalone = ( ALPHA / "__" / "_" / ":" ) *( ALPHA / DIGIT / "_" / ":" )
+
+metricname = ( metricname-namespace metricname-namespace-suffix metricname-suffix / metricname-standalone )
 
 labelset = ( "" / "{}" / labelset-labels)
 
 labelset-labels = { ( labelname "=" DQUOTE 1*UTF8-char DQUOTE) *("," labelname "=" DQUOTE 1*UTF8-char DQUOTE) }
 
-labelname = ALPHA *( ALPHA / DIGIT / "_" )
+labelname = ( ALPHA / "_" ) *( ALPHA / DIGIT / "_" / )
 
 # TODO https://github.com/OpenObservability/OpenMetrics/pull/93#discussion_r213579928
-datapoint = ( float / "+Inf" / "-Inf" / "NaN" )
+datapoint = ( float / "+Inf" / "-Inf" / "-Infinity" / "NaN" )
 
 float = 1*DIGIT *1( "." 1*DIGIT *1( "e" ( "+" / "-" ) 1*3DIGIT ) )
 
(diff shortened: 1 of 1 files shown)


draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,90 @@
+# Rationale
+
+The monitoring world is fragmented and there is no up-to-date standard.
+The closest to an universal monitoring standard we have is SNMP.
+While it is reasonably well-supported in the networking world, it failed to gain wide adoption beyond this field.
+SNMP also carries over many engineering trade-offs which made sense at the time, but are not matching well to today's trade-offs; a prime example would be the need to support ASN.1 with its many intricacies.
```
brian-brazil: I'd also mention the chatty nature of the protocol, which doesn't work for large data sets.

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,90 @@
...
+
+While there have been other efforts, none of them gained any wide-spread traction.
+Furthermore, most standardization attempts have been foiled by trying to establish themselves in a niche and expanding from there.
+
+More and more fields of IT are being swept up by commoditization, currently aka foo-as-code.
+HTTP endpoints are replacing TCP/IP-based services, and UTF-8 has established itself as the universal non-binary encoding.
+Finally, metrics-based monitoring is overhauling large parts of the monitoring landscape due to their scalability and adaptability.
+
+Given all this, OpenMetrics is an effort to establish a new and universal standard.
```
brian-brazil: ...for metrics-based monitoring.

We really need a bit here explaining what metrics are

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,90 @@
...
+There are two wire formats specified in OpenMetrics: The text format and the protobuf format.
+All client implementations MUST be able to emit text format; they MAY also emit protobuf if negotiated.
+All server implementations MUST be able to ingest text format; they MAY also ingest protobuf if negotiated.
+All implementations MUST default to the text format.
+
+TODO rationale why text is default
+
+# Wire formats
+
+Special care has been taken to make the wire formats efficient.
```
brian-brazil: This isn't the case. Easy to generate and understand is the main concern.

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,90 @@
...
+The three main considerations are CPU time, RAM usage, and network utilization, in this order.
+OpenMetrics is a type 3 grammar, making writing quick and small parsers possible.
+UTF-8 compresses well, and protobuf is already binary.
```
brian-brazil: How is protobuf being binary relevant?

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,90 @@
...
+
+## Text format
+
+Metric names MUST appear as a group.
+A group MAY have intial lines with metadata, starting with `# $keyword` and MUST have at least one line with metric data.
```
brian-brazil: It's fine to have metadata without a sample.

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,90 @@
...
+
+ABNF as per RFC 5234, UTF8-char as per RFC 3629.
+
+group = *metadata 1*metricdata
+
+metadata = # SP ("HELP" / "TYPE" / TODO) SP metadata-text CRLF
```
brian-brazil: Only LF, no CR

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,90 @@
...
+
+metadata-text = UTF8-char *(UTF8-char / SP) # TODO - how to do UTF-8?
+
+metricdata = metricname labelset SP datapoint *1(SP timestamp) *1exemplar CRLF
```
brian-brazil: exemplars are only possible in some cases

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,90 @@
...
+
+metricname = 1*(ALPHA / DIGIT / "_")
```
brian-brazil: colons are permitted. Metric names can't start with a digit.

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,90 @@
...
+UTF-8 compresses well, and protobuf is already binary.
+
+## Text format
+
+Metric names MUST appear as a group.
+A group MAY have intial lines with metadata, starting with `# $keyword` and MUST have at least one line with metric data.
+
+ABNF as per RFC 5234, UTF8-char as per RFC 3629.
+
+group = *metadata 1*metricdata
```
brian-brazil: Don't forget the EOF

draft-hartmann-openmetrics-00
RichiH: Not specced yet, but do we agree that they can't start with underscore, either?

draft-hartmann-openmetrics-00
RichiH: IIRC we allow them for any time series, but agreed that they don't make a lot of sense outside of histograms.

draft-hartmann-openmetrics-00
RichiH: Inasmuch it's already compressed.

draft-hartmann-openmetrics-00
RichiH: Done.

draft-hartmann-openmetrics-00
RichiH: Done.

draft-hartmann-openmetrics-00
RichiH: Adapted.

draft-hartmann-openmetrics-00
RichiH: True.

draft-hartmann-openmetrics-00
brian-brazil: The spec says that is permitted.

draft-hartmann-openmetrics-00
brian-brazil: The spec says only histograms

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,108 @@
...
+
+metricdata = metricname labelset SP datapoint *1(SP timestamp) *1exemplar LF
+
+metricname = ALPHA *(ALPHA / DIGIT / "_")
+
+labelset = ( "" / "{}" / labelset-labels)
+
+labelset-labels = { (1*UTF8-char "=" DQUOTE 1*UTF8-char DQUOTE) *("," 1*UTF8-char "=" DQUOTE 1*UTF8-char DQUOTE) }
+
+datapoint = 1*DIGIT *1("." *DIGIT *1( "e" ( "+" / "-" ) 2DIGIT ) )
```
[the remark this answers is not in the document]
brian-brazil: Float64 allows for 3 digits in the exponent. The spec says anything the Go float parser takes, or xxx.yyyyy.

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,108 @@
...
+
+## Text format
+
+Metric names MUST appear as a group.
+
+### ABNF
+
+ABNF as per RFC 5234, UTF8-char as per RFC 3629.
+
+group = *metadata 1*metricdata
```
brian-brazil: metricdata is optional

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,108 @@
...
+
+metadata = # SP ("HELP" / "TYPE" / TODO) SP metadata-text LF
+
+metadata-text = UTF8-char *(UTF8-char / SP) # TODO - how to do UTF-8?
+
+metricdata = metricname labelset SP datapoint *1(SP timestamp) *1exemplar LF
+
+metricname = ALPHA *(ALPHA / DIGIT / "_")
```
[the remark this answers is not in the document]
RichiH: What about _ ?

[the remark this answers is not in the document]
RichiH: Specifically single `_` as `__` is reserved. I would tend to forbid any leading `_` for now but noting that future use might allow while keeping it reserved.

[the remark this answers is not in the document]
brian-brazil: _ is not a digit, so it can be first.

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,108 @@
...
+
+labelset = ( "" / "{}" / labelset-labels)
+
+labelset-labels = { (1*UTF8-char "=" DQUOTE 1*UTF8-char DQUOTE) *("," 1*UTF8-char "=" DQUOTE 1*UTF8-char DQUOTE) }
+
+datapoint = 1*DIGIT *1("." *DIGIT *1( "e" ( "+" / "-" ) 2DIGIT ) )
+
+timestamp = 1*DIGIT # TODO adapt to timestamp discussion
```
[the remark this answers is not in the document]
RichiH: I thought we agreed not to over-complicate by restricting to `1*DIGIT *1("." 1*DIGIT)`, but I don't feel strongly either way.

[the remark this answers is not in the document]
brian-brazil: We didn't over complicate it by making all types of numbers the same. This is all extensively covered  in the text draft, please work from that.

draft-hartmann-openmetrics-00
RichiH: Extended a bit.

draft-hartmann-openmetrics-00
RichiH: There seem to be valid use cases, e.g. why a bool changed; I suggest we revisit this in our call.

draft-hartmann-openmetrics-00
RichiH: Are there use cases for this or did we simply not limit this enough?

brian-brazil: We're following what Prometheus currently does, and it can come up when sanitizing metrics

draft-hartmann-openmetrics-00
```diff
@@ -68,21 +68,25 @@ metadata-text = UTF8-char *(UTF8-char / SP)
...
 metricname = ( ALPHA / ":" ) *( ALPHA / DIGIT / "_" / ":" )
 
 labelset = ( "" / "{}" / labelset-labels)
 
+labelset-labels = { ( labelname "=" DQUOTE 1*UTF8-char DQUOTE) *("," labelname "=" DQUOTE 1*UTF8-char DQUOTE) }
+
 labelname = ALPHA *( ALPHA / DIGIT / "_" )
 
-labelset-labels = { ( labelname "=" DQUOTE 1*UTF8-char DQUOTE) *("," labelname "=" DQUOTE 1*UTF8-char DQUOTE) }
+# TODO https://github.com/OpenObservability/OpenMetrics/pull/93#discussion_r213579928
```
brian-brazil: There's a bit more than that now, as it turns out Java uses -Infinity. The draft test format has the details.

draft-hartmann-openmetrics-00
```diff
@@ -99,13 +103,22 @@ go_gc_duration_seconds_sum 9241.311909524
...
+```
+promhttp_metric_handler_requests_total{code="200"} 27205 1546777500
+promhttp_metric_handler_requests_total{code="200"} 27401 1546777666
+```
 
 ## protobuf format
 
 # Security Considerations
 
+Authentication, authorization, and accounting SHOULD be handled outside of OpenMetrics.
```
brian-brazil: That implies that you must have AAA

draft-hartmann-openmetrics-00
```diff
@@ -99,13 +103,22 @@ go_gc_duration_seconds_sum 9241.311909524
...
+#TODO should we require TLS 1.2 at a mimimum? # https://github.com/OpenObservability/OpenMetrics/issues/61
+All client implementations SHOULD be able to secure their HTTP traffic with TLS 1.2 or later.
```
brian-brazil: I'm not sure we should specify TLS versions, as those will change over time

draft-hartmann-openmetrics-00
brian-brazil: I don't see how that's a valid use case

draft-hartmann-openmetrics-00
RichiH: As to the former, this could be allowed in a compat mode. It's a chance to clean up a little.

As to the latter, I can only think of situations in which something's broken.

draft-hartmann-openmetrics-00
RichiH: As this is a minimum, I don't think there's a negative effect other than potentially excluding somehting which does not do TLS 1.2 yet.

draft-hartmann-openmetrics-00
brian-brazil: That's the world we live in, I don't see the need to introduce a need for a compatibility mode - this is valid input.

draft-hartmann-openmetrics-00
RichiH: Updated.

draft-hartmann-openmetrics-00
RichiH: So my worry is about having a min at all, not about anything else.

draft-hartmann-openmetrics-00
RichiH: Consensus from call: Only on histograms and gauge histograms

draft-hartmann-openmetrics-00
RichiH: metricnames can start with : _ ALPHA, but not with numbers.

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,161 @@
...
+
+## protobuf format
+
+# Security Considerations
+
+Implementors MAY choose to offer authentication, authorization, and accounting; if they so choose, this SHOULD be handled outside of OpenMetrics.
+#TODO should we require TLS 1.2 at a mimimum? # https://github.com/OpenObservability/OpenMetrics/issues/61
+All client implementations SHOULD be able to secure their HTTP traffic with TLS 1.2 or later.
+If a client implementation does not support encryption, operators SHOULD use reverse proxies, firewalling, and/or ACLs for AAA.
+All server implementations MUST be able to ingest data from endpoints secured with TLS 1.2 or later.
```
robskillington: As briefly talked about, s/MUST/SHOULD here I imagine? Also not sure what what this means in the case of push, as also discussed.

draft-hartmann-openmetrics-00
RichiH: I added this, but shouldn't we arguably try to force Java to not fall out of line?

draft-hartmann-openmetrics-00
RichiH: Are we expecting ingestors to be this resource-constrained?

Push is completely detached and will be handled in a separate doc.

draft-hartmann-openmetrics-00
brian-brazil: That'd make it harder for a basic Java implementation to be correct out of the box.

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,172 @@
...
+Metric names MUST appear as a group.
+If a metricname has appeared in a group before it MUST NOT be repeated after that group finishes.
+
+
+
+### ABNF
+
+ABNF as per RFC 5234, UTF8-char as per RFC 3629.
+
+data = *group LF EOF
```
brian-brazil: There's an optional (SHOULD) LF after the EOF.

Does EOF need to be defined? As written I'd take it as 0x1B rather than "EOF".

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,172 @@
...
+
+metricname-suffix = ( ALPHA / ":" ) *metricname-inline
+
+metricname-standalone = ( ALPHA / "__" / "_" / ":" ) *metricname-inline
+
+metricname = ( metricname-namespace metricname-namespace-suffix metricname-suffix / metricname-standalone )
+
+labelset = ( "" / "{}" / labelset-labels)
+
+labelset-labels = { ( labelname "=" DQUOTE 1*UTF8-char DQUOTE) *("," labelname "=" DQUOTE 1*UTF8-char DQUOTE) }
```
brian-brazil: There's no requirement that there be at least one character in a label value

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,172 @@
...
+
+labelname = ( ALPHA / "_" ) *( ALPHA / DIGIT / "_" / )
+
+# TODO https://github.com/OpenObservability/OpenMetrics/pull/93#discussion_r213579928
+datapoint = ( float / "+Inf" / "-Inf" / "-Infinity" / "NaN" )
+
+float = 1*DIGIT *1( "." 1*DIGIT *1( "e" ( "+" / "-" ) 1*3DIGIT ) )
```
brian-brazil: .1, 1., 1e1, and 1.0E+1 are all valid.

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,172 @@
...
+
+# TODO adapt to timestamp discussion
+timestamp = 1*DIGIT
```
brian-brazil: It's a float

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,172 @@
...
+
+# TODO put in length limit
+exemplar = "#" SP 1*(UTF8-char / SP)
```
brian-brazil: There's more to an exemplar than this

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,172 @@
...
+
+ABNF as per RFC 5234, UTF8-char as per RFC 3629.
+
+data = *group LF EOF
+
+group = *metadata *metricdata
+
+metadata = ( "" / # SP ( "HELP" / "TYPE" / "UNIT" ) SP metadata-text LF )
+
+metadata-text = *(UTF8-char / SP)
```
brian-brazil: How does this handle escaping?

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,172 @@
...
+
+# TODO https://github.com/OpenObservability/OpenMetrics/pull/93#discussion_r248005922
+metricdata = metricname labelset SP datapoint *1(SP timestamp) *1exemplar LF
+
+metricname-inline = ( ALPHA / DIGIT / ":" ) *( ALPHA / DIGIT / "_" / ":" )
```
brian-brazil: This doesn't seem right, and the grammar is ambiguous.

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,172 @@
...
+All implementations SHOULD use TLS to transmit data.
+
+# IANA Considerations
+
+While currently most implementations of the Prometheus exposition format are using non-registered ports[Prometheus ports], OpenMetrics can be found at a well-defined port.
+Metric exposition should be independent of production services; as such, ports like TCP/80, TCP/443, and TCP/8080 are not suitable for the use for OpenMetrics.
+The port assigend by IANA for clients exposing data is XXX.
+
+# TODO discuss this - https://github.com/OpenObservability/OpenMetrics/issues/123
+If more than one metric endpoint needs to be reachable at a common IP address, operators SHOULD consider using a reverse proxy to reach endpoints on localhost addresses and ports.
```
brian-brazil: That's not our standard deployment model, it's multiple ports. I think you're over extending the network-device case.

RichiH: I know, but I expect massive pushback from IETF unless we have a good story here.

brian-brazil: I'm not saying we shouldn't have a story, I'm saying we should not present this as the primary deployment model.

draft-hartmann-openmetrics-00
```diff
@@ -0,0 +1,172 @@
...
+
+metricname-namespace = [ ( ALPHA / "_" / ":" ) ] *metricname-inline ( ALPHA / DIGIT / ":" )
+
+metricname-namespace-suffix = "__"
+
+metricname-namespaced = ( ALPHA / ":" ) *metricname-inline
+
+metricname-standalone = ( ALPHA / "_" / ":" ) *metricname-inline
+
+metricname = ( [ metricname-namespace ] metricname-namespace-suffix metricname-namespaced / metricname-standalone )
```
brian-brazil: This will reject a_b_c_

brian-brazil:
This is obselete.

PR #94, opened 2018-07-19
Clarify names

pilhuhn:
The metric names syntax should not contain the colon, as folks see it and don't read further.


brian-brazil:
The current wording is correct, there are valid use cases for colons being seen in the format.

pilhuhn:
So can we list them? Perhaps the name section would need to be broken apart for
* data that goes over the wire from exporter to TSDB server
* queries to the TSDB server that may include colons

RichiH:
https://github.com/OpenObservability/OpenMetrics/blob/master/OpenMetrics.md is now published

PR #126, opened 2019-01-15
data model proto

RichiH:


proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+      repeated double bounds = 1;
+    }
+
+  }
+
+  // The number of values in each bucket of the histogram, as described in
+  // `bucket_options`.
+  //
+  // `bucket_counts` will typically contain N + 1 values. If you supply fewer
+  // values, the remaining values are assumed to be 0.
```
brian-brazil: This doesn't make sense, as these are cumulative. I'd suggest requiring it to be explicit

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+    Exemplar exemplar = 2;
+    message Exemplar {
+      // Required.
+      // The value of the example.
+      double value = 1;
+
+      // The timestamp of the example. Optional but SHOULD set it.
+      Timestamp timestamp = 2;
+
+      // Additional information about the example value // (e.g. trace id).
```
brian-brazil: Length limit should be mentioned

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+
+  // `BucketOptions` describes the bucket boundaries used to create a
+  // histogram. The buckets can be in a linear sequence, an
+  // exponential sequence, or each bucket can be specified explicitly.
+  // `BucketOptions` does not include the number of values in each bucket.
+  //
+  // A bucket has a 0 lower bound and an inclusive upper bound for the
+  // values that are counted for that bucket. That is, buckets are
+  // cumulative. The upper bound of successive buckets must be increasing.
+  // There is an implicit overflow bucket that extends up to +infinity.
```
brian-brazil: and must match the value of _count

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+  // interval.
+  Timestamp start_timestamp = 7;
+
+  // If not specified, the timestamp will be decided by the backend.
+  Timestamp timestamp = 8;
+}
+
+// Value for CUMULATIVE_HISTOGRAM or GAUGE_HISTOGRAM point.
+message HistogramValue {
+  // Required.
```
brian-brazil: Neither of these should be required

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+// Value for SUMMARY point.
+message SummaryValue {
+  // Sum and count are optional since some systems (e.g. DropWizard) do not
+  // expose them. The start_timestamp only applies to the sum and count.
+  // The quantiles can be reset at arbitrary unknown times.
+
+  double sum = 1;
+  int64 count = 2;
+
+  repeated Quantile quantile = 3;
```
brian-brazil: It should be clear that this is optional

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+message Point {
+  oneof value {
+    double double_value = 1;
+    int64 int_value = 2;
+    HistogramValue histogram_value = 3;
+    StateSetValue state_set_value = 4;
+    InfoValue info_value = 5;
+    SummaryValue summary_value = 6;
+  }
+  // Required for COUNTER, SUMMARY or CUMULATIVE_HISTOGRAM type.
```
brian-brazil: Optional

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+  //
+  // The cumulative value is over
+  // the time interval (start_timestamp, timestamp].
+  // If start_timestamp is present and timestamp is absent,
+  // the backend assigned timestamp may be used to construct the time
+  // interval.
+  Timestamp start_timestamp = 7;
+
+  // If not specified, the timestamp will be decided by the backend.
```
brian-brazil: Make clear this is optional, and discouraged

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+message MetricDescriptor {
+  // Required.
+  // Names must match the regex [a-zA-Z_:][a-zA-Z0-9_:]* however colons are
+  // reserved for monitoring system use, and names beginning with
+  // underscores are reserved for monitoring-system internal use.
+  string name = 1;
+
+  // TODO: Format?
+  // STATE_SETs and INFO types must not have a unit.
+  string unit = 2;
```
robskillington: Should this perhaps instead be an enum?

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+
+  // Required.
+  string value = 2;
+}
+
+message Timestamp {
+  // Required.
+  int64 seconds = 1;  // Seconds since the epoch. Negative values are permitted
+
+  uint32 nanoseconds = 2;
```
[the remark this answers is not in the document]
robskillington: `nanoseconds_offset` perhaps is a better concept?  And if its signed instead, perhaps that's more permissive and allows you to simply always just blindly add these two together:
```
unix_nanos = (timestamp.seconds * NANOSECONDS_IN_SECONDS) + timestamp.nanoseconds_offset
```

Thoughts?

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// A point in a timeseries.
+message Point {
+  oneof value {
+    double double_value = 1;
+    int64 int_value = 2;
```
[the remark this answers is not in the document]
robskillington: Would it be a little strange that double value can be negative but int value cannot be?

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+    // to the previous bucket (except the overflow bucket). This implies
+    // a constant absolute uncertainty on the specific value of a sample.
+    //
+    // There are `num_explicit_buckets` (N) buckets plus the overflow
+    // bucket. Bucket `i` (0 <= i < N) has upper bound : offset + (width * i).
+    // Note that when offset = 0, bucket 0 represents the interval [0, 0].
+    message Linear {
+      // Required.
+      // Must be greater than 0.
+      int32 num_explicit_buckets = 1;
```
robskillington: uint32 perhaps since there's no such thing as negative buckets?

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+    //
+    // There are `num_exponential_buckets` + 1 (N) buckets plus the overflow
+    // bucket.
+    // Bucket 0 represents the interval [0, 0].
+    // Bucket `i` (1 <= i < N) has upper bound :
+    //   scale * (growth_factor ^ (i - 1)).
+    message Exponential {
+      // Required.
+      // Must be greater than 0.
+      int32 num_exponential_buckets = 1;
```
robskillington: uint32 perhaps here too since there's no such thing as negative buckets?

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// Value for SUMMARY point.
+message SummaryValue {
+  // Sum and count are optional since some systems (e.g. DropWizard) do not
+  // expose them. The start_timestamp only applies to the sum and count.
+  // The quantiles can be reset at arbitrary unknown times.
+
+  double sum = 1;
+  int64 count = 2;
```
robskillington: uint64 since there can't be negative sampled values?

proto/openmetrics_data_model.proto
sumeer: why not?

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+  Timestamp timestamp = 8;
+}
+
+// Value for CUMULATIVE_HISTOGRAM or GAUGE_HISTOGRAM point.
+message HistogramValue {
+  // Required.
+  double sum = 1;
+
+  // Required.
+  int64 count = 2;
```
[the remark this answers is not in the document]
sumeer: done

proto/openmetrics_data_model.proto
sumeer: Done

proto/openmetrics_data_model.proto
sumeer: Done

proto/openmetrics_data_model.proto
sumeer: done

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+  // `bucket_counts` will typically contain N + 1 values. If you supply fewer
+  // values, the remaining values are assumed to be 0.
+  //
+  // The order of the values in `bucket_counts` follows the bucket numbering
+  // schemes described for the three bucket types.
+  repeated BucketCount bucket_counts = 4;
+  message BucketCount {
+    // Required.
+    // Count is the number of values for a bucket of the histogram.
+    int64 count = 1;
```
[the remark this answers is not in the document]
sumeer: done

proto/openmetrics_data_model.proto
sumeer: done

proto/openmetrics_data_model.proto
sumeer: are you saying the quantile's are optional?

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,270 @@
...
+  // expose them. The start_timestamp only applies to the sum and count.
+  // The quantiles can be reset at arbitrary unknown times.
+
+  double sum = 1;
+  int64 count = 2;
+
+  repeated Quantile quantile = 3;
+  message Quantile {
+    // Required.
+    // Must be in the interval (0.0, 1.0].
```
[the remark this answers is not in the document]
sumeer: how do you calculate the 0th percentile from a sample of values? i don't think subtracting delta from the lowest value in the sample is mathematically sound.

proto/openmetrics_data_model.proto
sumeer: Making it consistent with "The maximum number of characters in the labels, including the curly braces, is 64 UTF-8 characters." seems messy. I'd rather define a length limit on the strings in the Label and compute the resulting limit for the text format (accounting for the quirks in the text format in the text format specification)

proto/openmetrics_data_model.proto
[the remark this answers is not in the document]
brian-brazil: Same question for 100th percentile. It's either both or neither.

proto/openmetrics_data_model.proto
brian-brazil: Yes

proto/openmetrics_data_model.proto
brian-brazil: Some systems don't have them (MySQL)

proto/openmetrics_data_model.proto
sumeer: these are just the bucket boundaries. I've added a comment later about the sum of bucket counts.

proto/openmetrics_data_model.proto
brian-brazil: No, we had not agreed this. It's optional, as many many systems out there do not support this.

proto/openmetrics_data_model.proto
brian-brazil: I wouldn't say encouraged for push, it depends on what's on the other side. The other side can always add its own.

proto/openmetrics_data_model.proto
brian-brazil: It'd have to be the same limits as metric names.

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,278 @@
...
+// one or more points.
+message MetricPoints {
+  // Required.
+  MetricDescriptor metric_descriptor = 1;
+
+  repeated Timeseries timeseries = 2;
+}
+
+// A single timeseries.
+message Timeseries {
```
bogdandrutu: I don't see anywhere something like `MonitoredResource` in Stackdriver or `__meta__` labels (in Prometheus, hope I used the correct name). Is this intentional? How are these reported?

Couple of use-cases where these are necessary and cannot be determine even if this data model is focused on a pull mechanism which would allow the backend to annotate these meta labels in general:
1. Task (metric producer) -> Proxy -> Metric Backend. In this case the meta labels should be included in the metric because the backend does not know about the metric producer and cannot associate the right meta labels.

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,278 @@
...
+// timeseries, value of INFO metrics, and exemplars in Histograms.
+message Label {
+  // Required.
+  string name = 1;
+
+  // Required.
+  string value = 2;
+}
+
+message Timestamp {
```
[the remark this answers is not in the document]
brian-brazil: I think it'd be better to keep self-contained

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,278 @@
...
+
+// This is a description of the OpenMetrics data model in a proto
+// form. Additionally, it serves as the proto wire format.
+//
+// Interpret "required" as semantically required. Some fields are
+// semantically required depending on the values of the other fields
+// (and indicated by comments).
+
+// The top-level message sent on the wire.
+message MetricSet {
```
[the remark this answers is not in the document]
brian-brazil: It's a set, you aren't allowed have duplicates.

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,278 @@
...
+  string unit = 2;
+
+  // Required.
+  Type type = 3;
+
+  string description = 4;
+
+  enum Type {
+    UNKNOWN = 0;  // double or int valued point
+    GAUGE = 1;    // double or int valued point
```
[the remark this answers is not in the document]
brian-brazil: We discussed this at length, this is the way we're going.

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,278 @@
...
+  // changes in bucketing for a metric. Same for SUMMARY (quantiles are
+  // not defined here), and STATE_SET (the set of possible values is
+  // defined in each point).
+}
+
+// A name-value pair. These are used in multiple places: identifying
+// timeseries, value of INFO metrics, and exemplars in Histograms.
+message Label {
+  // Required.
+  string name = 1;
```
bogdandrutu: Do we want to support a label description? Human readable description of this metric?

proto/openmetrics_data_model.proto
brian-brazil: This is what Info metrics are for.

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,278 @@
...
+      // Required.
+      // The value of the example.
+      double value = 1;
+
+      // The timestamp of the example. Optional but SHOULD set it.
+      Timestamp timestamp = 2;
+
+      // Additional information about the example value // (e.g. trace id).
+      // The sum of lengths of all the strings in all the labels must not
+      // exceed 64 UTF-8 characters.
```
bogdandrutu: 64 characters for all the labels here seems unreasonable. trace-id has 32-hex characters and span-id has 16-hex characters only these values will represent almost the entire limit.

See https://github.com/w3c/trace-context

proto/openmetrics_data_model.proto
bogdandrutu: Nice, but I have some problems understanding some of the things here, maybe worth some clarification:
1. I don't understand how does the correlation between Info metrics and the other metrics happen. Should we have one info metric per MetricSet? Can I have in the same MetricSet metrics from multiple producers (in case of a proxy if I monitor multiple tasks)?

proto/openmetrics_data_model.proto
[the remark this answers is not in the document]
bogdandrutu: At least we can copy the definition from there and avoid any arguments about nanoseconds being int vs uint.

proto/openmetrics_data_model.proto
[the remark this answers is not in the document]
bogdandrutu: Then probably document that, to make it clear why is a set and what is the "hash" function.

proto/openmetrics_data_model.proto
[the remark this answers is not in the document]
bogdandrutu: You mean the current state or what I proposed?

proto/openmetrics_data_model.proto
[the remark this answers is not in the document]
sumeer: do you know why it is int32 in google/protobuf/timestamp.proto even though it says "Non-negative fractions of a second at nanosecond resolution."?

proto/openmetrics_data_model.proto
sumeer: There is some concern about allowing this to be a dumping ground for stuff.
With UTF-8, the current limit lets someone use 64*4 = 256 bytes. 
How about we state the limit in the sum of bytes needed for names and values so that it doesn't penalize users who are restricting to ascii.

proto/openmetrics_data_model.proto
brian-brazil: We discussed this previously, and no.

proto/openmetrics_data_model.proto
brian-brazil: The spec in general is UTF-8, I'd rather not switch to bytes here.

It sounds to me like you've 12 characters left over for the label name, so you're inside the limit.

proto/openmetrics_data_model.proto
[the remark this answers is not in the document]
bogdandrutu: Two reasons that I know:
1. This needs to be compatible with languages that do not have unsigned (like Java).
2. It makes diff calculation easier (you do the diff initially then normalize the nanos).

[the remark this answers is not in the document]
brian-brazil: I don't think either is a problem in practice, it's not going to have values where that'll be an issue.

proto/openmetrics_data_model.proto
bogdandrutu: @sumeer I am not suggesting to not have a limit, but the limit is too small.

@brian-brazil correct but if I use "trace-id" and "span-id" (32 + 8 + 16 + 7 = 63) as keys then I have one character left so no other exemplar. Also we would like to have a trace-options which will suggest if the trace is sample or not (2-hex characters).

I know that the solution can be to modify the w3c standard (impossible) or to use "shorter" keys like "tid" and "sid" and "to" but those are not that clear. I really think that the limit should be more realistic and we can go with 256 characters for this.

proto/openmetrics_data_model.proto
bogdandrutu: roger that.

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,278 @@
...
+  // Note: No bucketing is defined for a HISTOGRAM in the descriptor.
+  // It is an implementation detail of the backend whether it can handle
+  // changes in bucketing for a metric. Same for SUMMARY (quantiles are
+  // not defined here), and STATE_SET (the set of possible values is
+  // defined in each point).
+}
+
+// A name-value pair. These are used in multiple places: identifying
+// timeseries, value of INFO metrics, and exemplars in Histograms.
+message Label {
```
bogdandrutu: Because label is key, value you can use Labels as map<String,String>.

proto/openmetrics_data_model.proto
sumeer: I think we are allowing multiple info metrics, e.g. build_info. It is up to the receiver to decide what to do with them. But they will all apply to the whole MetricSet, so one shouldn't mix multiple producers in the same MetricSet.

brian-brazil: Yes, there's no limit on how many info metrics you can have.

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,278 @@
...
+message StateSetValue {
+  // Each state must have a unique name. Any number of states may be enabled.
+  repeated State states = 1;
+
+  message State {
+    // Required.
+    bool enabled = 1;
+
+    // Required.
+    string name = 2;
```
bogdandrutu: Why is this not a regular label? Recommended key can be `state_name` the label value is the `state` then this becomes a single boolean.

brian-brazil: Then it'd not be a first-class type.

bogdandrutu: I am fine not having this as a first class. Probably on the gauge supporting boolean is enough

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,278 @@
...
+
+  // Required.
+  BucketOptions bucket_options = 3;
+
+  // `BucketOptions` describes the bucket boundaries used to create a
+  // histogram. The buckets can be in a linear sequence, an
+  // exponential sequence, or each bucket can be specified explicitly.
+  // `BucketOptions` does not include the number of values in each bucket.
+  //
+  // A bucket has a 0 lower bound and an inclusive upper bound for the
```
brian-brazil: This should say "inclusive 0 lower bound" for clarity

beorn7: Buckets for values <0 don't seem possible with this data model. Is that intended?

brian-brazil: Per our discussions it is, noone could come up with a realistic use case.

beorn7: My usual tech use-case is Spam Assassin scores.

However, OpenMetrics is supposed to be generally applicable. Observing negative values is a very normal thing for histograms in general. It is almost zero cost to support it. Kicking the support out would be a serious blow to the applicability of the format.

beorn7: Current handling of negative bounds by Prometheus is defined here: https://prometheus.io/docs/prometheus/latest/querying/functions/#histogram_quantile

proto/openmetrics_data_model.proto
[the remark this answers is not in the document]
robskillington: As discussed we'll just use the google protobuf timestamp type here.

proto/openmetrics_data_model.proto
```diff
@@ -0,0 +1,292 @@
...
+  // Optional.
+  // Labels applying to all points, must be unique.
+  repeated Label labels = 1;
+
+  // Optional.
+  // If there is more than one point, all must have a timestamp field
+  // and they must be in increasing order of timestamp.
+  // If there is only one point and has no timestamp field, the receiver
+  // will assign a timestamp.
+  repeated Point points = 2;
```
bogdandrutu, line 83: The amount of cases you have multiple points vs one point for the same labels is extremely small, I would suggest to optimize for the case where only one point per labels is sent and remove repeated (avoids an allocation or two depends on the languages) with the downside of repeating the labels for the case where multiple points are needed.

proto/openmetrics_data_model.proto
bogdandrutu: This comment should be marked as resolved.

robskillington (approved):
LGTM, after merging latest changes now merging this to master

RichiH:
Noice

PR #131, opened 2019-02-26
Initial simple openmetricstest tool implementation

robskillington:
Really basic but an easy way to test whether parsing fails/passes for test cases in a testdata directory.

Here's an example testing `true` and `false`:
```sh
# Compile from repository root
make openmetricstest

# Run and since echo will always accept any stdin observe success
./bin/openmetricstest -testdata-dir ./tests/testdata/parsers/simple_counter -cmd-parser-text true
2019/04/09 21:25:56 RUN test: simple_counter
2019/04/09 21:25:56 parse-result-validator ok
2019/04/09 21:25:56 PASS test: simple_counter
2019/04/09 21:25:56 OK passed=1, failed=0, total_failures=0

# Run and since the script will always fail regardless of stdin observe failure
./bin/openmetricstest -testdata-dir ./tests/testdata/parsers/simple_counter -cmd-parser-text false
2019/04/09 21:40:30 RUN test: simple_counter
2019/04/09 21:40:30 parse-result-validator error:
2019/04/09 21:40:30 > parse error: exit status 1
2019/04/09 21:40:30 FAIL test: simple_counter
2019/04/09 21:40:30 FAILED passed=0, failed=1, total_failures=1
```


tests/openmetricstest/main.go
```diff
@@ -0,0 +1,146 @@
...
+	"path"
+	"strings"
+)
+
+const (
+	testFileName = "test.json"
+)
+
+var (
+	testDataDirArg       = flag.String("testdata_dir", "testdata", "testdata directory to use")
```
brian-brazil: Hyphens and periods are what we usually use

tests/openmetricstest/main.go
```diff
@@ -0,0 +1,146 @@
...
+	cmdTestParserTextArg = flag.String("cmd_test_parser_text", "", "command to run to test parser in text mode")
+)
+
+func main() {
+	flag.Parse()
+
+	testDataDir := strings.TrimSpace(*testDataDirArg)
```
brian-brazil: We don't usually do this, if the user has a space we should use it

tests/openmetricstest/main.go
```diff
@@ -0,0 +1,146 @@
...
+		log.Fatalf("cannot read testdata dir: %v", err)
+	}
+
+	for _, file := range files {
+		if file.Name() == testFileName {
+			runTest(dir, opts)
+			continue
+		}
+		if file.IsDir() {
+			// Recurse
```
brian-brazil: You would use `filepath.Walk` for this

tests/openmetricstest/main.go
```diff
@@ -0,0 +1,146 @@
...
+			runTests(path.Join(dir, file.Name()), opts)
+		}
+	}
+}
+
+type testDef struct {
+	Parse   []parseDef   `json:"parse"`
```
brian-brazil: This seems overly nested, I'd put fold it into one struct

tests/openmetricstest/main.go
```diff
@@ -0,0 +1,146 @@
...
+			if err := cmd.Start(); err != nil {
+				log.Fatalf("cannot start cmd: %v", err)
+			}
+
+			if _, err := stdin.Write(input); err != nil {
+				log.Fatalf("cannot write input: %v", err)
+			}
+
+			if err := validateTestOutcome(test, cmd); err != nil {
+				log.Fatalf("failed test: test=%s, err=%v", dir, err)
```
brian-brazil: Might it be better to collect all the failures, so you don't have to get them one at a time?

tests/testdata/parsers/simple_counter/test.json
```diff
@@ -0,0 +1,15 @@
+{
+  "parse": [
+    {
+      "type": "text",
+      "file": "metrics.txt"
```
brian-brazil: I was thinking more that we'd inline the metrics here

tests/openmetricstest/main.go
robskillington: Sounds good, will use that.

tests/openmetricstest/main.go
robskillington: Sounds good, will remove the trim.

tests/openmetricstest/main.go
robskillington: Good call, will use idiomatic filepath.Walk instead.

tests/openmetricstest/main.go
robskillington: Fair, we can always add more later if we want multi-input and multi-result-checking support.

tests/openmetricstest/main.go
robskillington: Agreed, will update.

tests/testdata/parsers/simple_counter/test.json
robskillington: Let's discuss this tomorrow, the issue becomes that for proto file contents we won't be able to put binary data inline.  

Also for testing edge cases with weird cases, it's probably better not to go through a JSON parse step which could do some strange translation before running the test case I would think.

robskillington:
Address feedback, one comment still to discuss was my response to the request to inline the text for the test, here was my response:
> Perhaps we should discuss this during the biweekly call, the issue is that for testing edge cases with strange characters, it's probably better not to go through a JSON parse step which could do some strange translation before running the test case I would think. 

> Also for testing proto contents we won't be able to put binary data inline.

robskillington:
@brian-brazil this is good for another review.

I was wondering what CI we should add to this repository, CNCF has some CI support I had heard - should I investigate that?  (Just for purely making sure these tools builds and passes its own tests, etc).

brian-brazil:
CircleCI would be best I think.

robskillington:
Okay great, @brian-brazil I added a makefile and Circle CI config - I don't have access to add the repository to Circle CI though.

If someone could go through adding the repository to Circle CI then it technically should be able to run on this PR once it's ready.

src/cmd/openmetricstest/README.md
```diff
@@ -0,0 +1,67 @@
...
+2019/04/09 21:25:56 RUN test: simple_counter
+2019/04/09 21:25:56 parse-result-validator ok
+2019/04/09 21:25:56 PASS test: simple_counter
+2019/04/09 21:25:56 OK passed=1, failed=0, total_failures=0
+```
+
+### Testing text parsing with a script that always fails
+
+```
+# Create a shell script that will always fail on any stdin
```
brian-brazil: `false` would be simpler.

src/cmd/openmetricstest/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+```
+Usage of ./openmetricstest:
+  -cmd-parser-text string
+        command to run to test parser in text mode
+  -testdata-dir string
+        testdata directory to use (default "./tests/testdata")
+```
+
+## Testing an implementation
```
brian-brazil: It could be clearer what the contract is, that we'll pass the input to a binary and it should succeed/fail accordingly.

src/cmd/openmetricstest/main.go
```diff
@@ -0,0 +1,211 @@
...
+	for _, elem := range r.tests {
+		if len(elem.failures) == 0 {
+			success++
+		} else {
+			failed++
+		}
+		failures += len(elem.failures)
+	}
+
+	result := fmt.Sprintf("passed=%d, failed=%d, total_failures=%d\n",
```
brian-brazil: This looks like a double `\n`

src/cmd/openmetricstest/main.go
```diff
@@ -0,0 +1,211 @@
...
+
+		testDir, _ := path.Split(filePath)
+		result, err := runTest(testDir, opts)
+		results.tests = append(results.tests, result)
+		return err
+	})
+	return results, err
+}
+
+type testDef struct {
```
brian-brazil: This can be flattened more.

src/cmd/openmetricstest/README.md
robskillington: Good point, I'll update to use `false` instead.

src/cmd/openmetricstest/README.md
robskillington: Good point, I'll update that.

src/cmd/openmetricstest/main.go
robskillington: Good catch, thanks.

robskillington:
@SuperQ if you don't mind weighing in on the basic Circle CI config `.circleci/config.yml` and the use of `Makefile` that'd be greatly appreciated!

src/cmd/openmetricstest/main.go
robskillington: Sounds good.

src/cmd/openmetricstest/main.go
```diff
@@ -102,26 +103,18 @@ func runTests(dir string, opts runTestsOptions) (runTestsResults, error) {
...
-
-type outcomeDef struct {
-	ParseResult *parseResultDef `json:"parseResult"`
+	Type        string          `json:"type"`
+	File        string          `json:"file"`
+	ResultParse *resultParseDef `json:"resultParse"`
 }
 
-type parseResultDef struct {
+type resultParseDef struct {
```
brian-brazil: Why not put this directly in the outcomeDef?

src/cmd/openmetricstest/main.go
```diff
@@ -102,26 +103,18 @@ func runTests(dir string, opts runTestsOptions) (runTestsResults, error) {
...
 	Valid bool `json:"valid"`
```
brian-brazil: shouldParse would be clearer

Makefile
```diff
@@ -0,0 +1,46 @@
+BUILD := $(abspath ./bin)
+BINARIES :=         \
+	openmetricstest \
+
+# To test "echo": 
+# make test-impl cmd-parser-text=echo
+# To test failure of "false":
```
brian-brazil: I'm unclear why there's a fail parameter

robskillington: Hey, this is so we can ensure that implementations we expect to fail do fail. (i.e. inverse the result).

This was mainly for CI, although I could try and use `bash -c "! make test-impl cmd-parser-text=false"` and see if that returns success on failure instead?

robskillington: I'll remove it and once CircleCI is working see if we can't just do something like I suggested.

src/cmd/openmetricstest/main.go
robskillington: Sure thing.

src/cmd/openmetricstest/main.go
robskillington: Sure, it's just in the future I wanted to support adding multiple types of verification (parsing just being one kind, exposition/serialization being another).

I'll just remove it if we don't want support for multiple types of result verifiers for now.

robskillington:
@brian-brazil this is ready for another round, ty.

src/cmd/openmetricstest/main.go
```diff
@@ -103,18 +103,9 @@ func runTests(dir string, opts runTestsOptions) (runTestsResults, error) {
...
-type resultParseDef struct {
-	Valid bool `json:"valid"`
-}
-
-// validator returns the concrete validator for this test result definition
-func (d resultParseDef) validator() testResultValidator {
-	return resultParseValidator{def: d}
+	Type        string `json:"type"`
+	File        string `json:"file"`
+	ShouldParse *bool  `json:"shouldParse"`
```
brian-brazil: Why the * ?

src/cmd/openmetricstest/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+This command specifies the process to run to test parsing OpenMetrics text exposition test cases.  It should take the OpenMetrics text exposition as stdin.
+
+## Examples
+
+These examples are from running the tool in the repository root after building it.
+
+### Testing text parsing with "echo"
+
+This is an example of tersting that "echo" can parse a simple counter text exposition:
```
brian-brazil: testing

true might be a better choice

src/cmd/openmetricstest/main.go
robskillington: Mainly so I know whether to run the parse validator or not, ideally in the future we have other validators and whether I should run them depends on whether the test definition has set the option or not.

src/cmd/openmetricstest/README.md
robskillington: Will update to `true` and thanks for the spelling catch.

robskillington:
@brian-brazil ready for another round.

src/cmd/openmetricstest/main.go
brian-brazil: That seems a bit overengineered for what we're currently doing, tristates can be confusing. I'd say keep it simple for now.

robskillington:
@brian-brazil addressed pointer comment.

PR #142, opened 2019-10-20
Protobuf updates, use protobuf.Timestamp for timestamps, cleanup comments and add explicit advice

robskillington:
These are updates based on the feedback provided by Bogdan and others and more closely matches the latest Protobuf format section of the draft RFC.

### DIFF
--- a/proto/openmetrics_data_model.proto
+++ b/proto/openmetrics_data_model.proto
@@ -1,17 +1,16 @@
 syntax = "proto3";
 
+// The OpenMetrics protobuf schema which defines the protobuf
+// wire format. 
+// Ensure to intepret "required" as semantically required for a valid message.
+// All string fields MUST be UTF-8 encoded strings.
 package openmetrics;
 
-// This is a description of the OpenMetrics data model in a proto
-// form. Additionally, it serves as the proto wire format.
-//
-// Interpret "required" as semantically required. Some fields are
-// semantically required depending on the values of the other fields
-// (and indicated by comments).
+import "google/protobuf/timestamp.proto";
 
-// The top-level message sent on the wire.
+// The top-level container type that is encoded and sent over the wire.
 message MetricSet {
-  // Each MetricPoints has one or more points from a single metric.
+  // Each MetricPoints has one or more points for a single metric.
   repeated MetricPoints metric_points = 1;
 }
 
@@ -44,32 +43,43 @@ message MetricDescriptor {
   // underscores are reserved for monitoring-system internal use.
   string name = 1;
 
+  // Required.
+  MetricType MetricType = 2;
+
   // STATE_SETs and INFO types must not have a unit.
   // Must match the regexp [a-zA-Z_:][a-zA-Z0-9_:]*
-  string unit = 2;
+  string unit = 3;
 
-  // Required.
-  Type type = 3;
-
+  // Optional description of the metric name.
   string description = 4;
 
-  enum Type {
-    UNKNOWN = 0;  // double or int valued point
-    GAUGE = 1;    // double or int valued point
-    COUNTER = 2;  // double or int valued point.
-    STATE_SET = 3;
-    INFO = 4;  // provides information about the entity being monitored
-    CUMULATIVE_HISTOGRAM = 5;
-    GAUGE_HISTOGRAM = 6;
-    // Summary quantiles are not recommended, since they cannot be
-    // aggregated.
-    SUMMARY = 7;
-  }
   // Note: No bucketing is defined for a HISTOGRAM in the descriptor.
   // It is an implementation detail of the backend whether it can handle
   // changes in bucketing for a metric. Same for SUMMARY (quantiles are
   // not defined here), and STATE_SET (the set of possible values is
   // defined in each point).
+}
+
+// The type of a metric.
+enum MetricType {
+  // Unknown must use double or int valued points.
+  UNKNOWN = 0;
+  // Gauge must use double or int valued points.
+  GAUGE = 1;
+  // Counter must use double or int valued points.
+  COUNTER = 2;
+  // State set must use state set value points.
+  STATE_SET = 3;
+  // Info provides information about the target being monitored
+  // and must use info value points.
+  INFO = 4;
+  // Cumulative histogram must use histogram value points.
+  CUMULATIVE_HISTOGRAM = 5;
+  // Gauge histogram must use histogram value points.
+  GAUGE_HISTOGRAM = 6;
+  // Summary quantiles are not recommended, since they cannot be
+  // aggregated.
+  SUMMARY = 7;
 }
 
 // A name-value pair. These are used in multiple places: identifying
@@ -82,15 +92,9 @@ message Label {
   string value = 2;
 }
 
-message Timestamp {
-  // Required.
-  int64 seconds = 1;  // Seconds since the epoch. Negative values are permitted
-
-  uint32 nanoseconds = 2;
-}
-
 // A point in a timeseries.
 message Point {
+  // Required.
   oneof value {
     double double_value = 1;
     int64 int_value = 2;
@@ -103,15 +107,16 @@ message Point {
   // Not required, but SHOULD be specified for COUNTER, SUMMARY or
   // CUMULATIVE_HISTOGRAM type.
   //
-  // The cumulative value is over
-  // the time interval (start_timestamp, timestamp].
+  // The cumulative value is over the time interval 
+  // (start_timestamp, timestamp].
   // If start_timestamp is present and timestamp is absent,
   // the backend assigned timestamp may be used to construct the time
   // interval.
-  Timestamp start_timestamp = 7;
+  google.protobuf.Timestamp start_timestamp = 7;
 
+  // Not required.
   // If not specified, the timestamp will be decided by the backend.
-  Timestamp timestamp = 8;
+  google.protobuf.Timestamp timestamp = 8;
 }
 
 // Value for CUMULATIVE_HISTOGRAM or GAUGE_HISTOGRAM point.
@@ -130,7 +135,7 @@ message HistogramValue {
   // exponential sequence, or each bucket can be specified explicitly.
   // `BucketOptions` does not include the number of values in each bucket.
   //
-  // A bucket has a 0 lower bound and an inclusive upper bound for the
+  // A bucket has an exclusive lower bound and an inclusive upper bound for the
   // values that are counted for that bucket. That is, buckets are
   // cumulative. The upper bound of successive buckets must be increasing.
   // There is an implicit overflow bucket that extends up to +infinity.
@@ -153,7 +158,7 @@ message HistogramValue {
     //
     // There are `num_explicit_buckets` (N) buckets plus the overflow
     // bucket. Bucket `i` (0 <= i < N) has upper bound : offset + (width * i).
-    // Note that when offset = 0, bucket 0 represents the interval [0, 0].
+    // Note that when offset = 0, bucket 0 represents the interval (0, 0].
     message Linear {
       // Required.
       // Must be greater than 0.
@@ -175,7 +180,7 @@ message HistogramValue {
     //
     // There are `num_exponential_buckets` + 1 (N) buckets plus the overflow
     // bucket.
-    // Bucket 0 represents the interval [0, 0].
+    // Bucket 0 represents the interval (0, 0].
     // Bucket `i` (1 <= i < N) has upper bound :
     //   scale * (growth_factor ^ (i - 1)).
     message Exponential {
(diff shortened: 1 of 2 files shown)


proto/openmetrics_data_model.proto
```diff
@@ -130,7 +135,7 @@ message HistogramValue {
   // exponential sequence, or each bucket can be specified explicitly.
   // `BucketOptions` does not include the number of values in each bucket.
   //
-  // A bucket has a 0 lower bound and an inclusive upper bound for the
+  // A bucket has an exclusive lower bound and an inclusive upper bound for the
```
brian-brazil: The lower bound is 0 inclusive.

proto/openmetrics_data_model.proto
```diff
@@ -153,7 +158,7 @@ message HistogramValue {
     //
     // There are `num_explicit_buckets` (N) buckets plus the overflow
     // bucket. Bucket `i` (0 <= i < N) has upper bound : offset + (width * i).
-    // Note that when offset = 0, bucket 0 represents the interval [0, 0].
+    // Note that when offset = 0, bucket 0 represents the interval (0, 0].
```
brian-brazil: [ was correct

proto/openmetrics_data_model.proto
```diff
@@ -21,11 +20,13 @@ message MetricPoints {
   // Required.
   MetricDescriptor metric_descriptor = 1;
 
+  // Required, at least one.
```
brian-brazil: Zero is fine

proto/openmetrics_data_model.proto
```diff
@@ -44,34 +45,45 @@ message MetricDescriptor {
...
   // defined in each point).
 }
 
+// The type of a metric.
+enum MetricType {
+  // Unknown must use double point values.
+  UNKNOWN = 0;
+  // Gauge must use double point values.
+  GAUGE = 1;
+  // Counter must use double point values.
```
brian-brazil: ints are fine

proto/openmetrics_data_model.proto
```diff
@@ -44,34 +45,45 @@ message MetricDescriptor {
...
+  // State set must use state set value points.
+  STATE_SET = 3;
+  // Info provides information about the target being monitored
+  // and must use info value points.
+  INFO = 4;
+  // Cumulative histogram must use histogram value points.
+  CUMULATIVE_HISTOGRAM = 5;
+  // Gauge histogram must use histogram value points.
+  GAUGE_HISTOGRAM = 6;
+  // Summary quantiles are not recommended, since they cannot be
```
brian-brazil: This doesn't really belong here, and would also seem to discourage quantileless summaryies.

proto/openmetrics_data_model.proto
```diff
@@ -44,34 +45,45 @@ message MetricDescriptor {
...
+enum MetricType {
+  // Unknown must use double point values.
+  UNKNOWN = 0;
+  // Gauge must use double point values.
+  GAUGE = 1;
+  // Counter must use double point values.
+  COUNTER = 2;
+  // State set must use state set value points.
+  STATE_SET = 3;
+  // Info provides information about the target being monitored
```
brian-brazil: Not always, it could be a subset of the target's information.

proto/openmetrics_data_model.proto
```diff
@@ -252,27 +264,29 @@ message StateSetValue {
...
+  // Optional, but SHOULD be specified. 
+  // This is for compatibility since some common systems do not expose them.
   double sum = 1;
+
+  // Optional, but SHOULD be specified. 
+  // This is for compatibility since some common systems do not expose them.
   uint64 count = 2;
 
-  // Not required, but SHOULD be specified.
+  // Required, at least one.
```
brian-brazil: These are optional.

proto/openmetrics_data_model.proto
```diff
@@ -192,16 +198,16 @@ message HistogramValue {
...
     }
 
-    // Specifies a set of buckets with arbitrary upper-bounds
+    // Specifies a set of buckets with arbitrary upper-bounds.
     //
     // There are `size(bounds)` (N) explicit buckets plus the overflow bucket.
     // Bucket `i` (0 <= i < N) has upper bound: bounds[i].
     // The `bounds` field must contain at least one element.
     message Explicit {
+      // Required, at least one.
```
brian-brazil: Zero is fine (though not very useful).

proto/openmetrics_data_model.proto
```diff
@@ -213,22 +219,27 @@ message HistogramValue {
...
+
     message Exemplar {
       // Required.
       // The value of the example.
       double value = 1;
 
-      // The timestamp of the example. Optional but SHOULD set it.
-      Timestamp timestamp = 2;
+      // Optional, but SHOULD set it.
+      // The timestamp of the example.
```
brian-brazil: exemplar

proto/openmetrics_data_model.proto
```diff
@@ -252,27 +264,29 @@ message StateSetValue {
 
 // Value for INFO point.
 message InfoValue {
+  // Required, at least one.
```
brian-brazil: Zero is fine, if not very useful.

proto/openmetrics_data_model.proto
```diff
@@ -175,8 +181,8 @@ message HistogramValue {
     //
     // There are `num_exponential_buckets` + 1 (N) buckets plus the overflow
     // bucket.
-    // Bucket 0 represents the interval [0, 0].
-    // Bucket `i` (1 <= i < N) has upper bound :
+    // Bucket 0 represents the interval (0, 0].
```
beorn7: This bucket would never receive any observations. Is that intended?

brian-brazil:
> Note that none of the histogram bucketing schemas enables negative observations.

Explicit does, linear should be fixed to.

proto/openmetrics_data_model.proto
```diff
@@ -192,16 +198,16 @@ message HistogramValue {
...
-    // Specifies a set of buckets with arbitrary upper-bounds
+    // Specifies a set of buckets with arbitrary upper-bounds.
     //
     // There are `size(bounds)` (N) explicit buckets plus the overflow bucket.
     // Bucket `i` (0 <= i < N) has upper bound: bounds[i].
     // The `bounds` field must contain at least one element.
     message Explicit {
+      // Required, at least one.
       // The values must be monotonically increasing and >= 0.
       repeated double bounds = 1;
```
beorn7: @brian-brazil This says that bounds must be >= 0. While that doesn't strictly prevent negative observations, all of them would always end up in bucket 0.

brian-brazil: Ah, I'd missed that. That needs updating then.

robskillington: Ok good call, I'll update that.

proto/openmetrics_data_model.proto
robskillington: Will restore int values (is interesting though, text format has no notion of distinguising between float64 vs ints, only protobuf will have the ability to do this).

proto/openmetrics_data_model.proto
robskillington: Will mark optional.

proto/openmetrics_data_model.proto
robskillington: Will restore int values (is interesting though, text format has no notion of distinguising between float64 vs ints, only protobuf will have the ability to do this).

proto/openmetrics_data_model.proto
robskillington: Will remove.

proto/openmetrics_data_model.proto
beorn7: Wait a sec, counters are floats, aren't they?

brian-brazil: Counters can be ints or floats.

beorn7: That makes sense.

Which makes me think that Gauges and Unknowns could also meaningfully be integers (bytes of data, number of goroutines, …). Is it deliberate that those _must_ be floats?

And @robskillington , IIRC, the text format marks integers by having neither a dot nor an `e` anywhere, while floats need to have at least one of them in their text representation. (We went to extra length to get that right in the Go client.)

brian-brazil: The use case for an int64 is counters for things like 100Gb network interfaces, where the potential loss of precision is important. If it's just a gauge, or you don't know, then a float is sufficient. This is all data model.

proto/openmetrics_data_model.proto
robskillington: It is inclusive of the zero value in the upper bound, hence values that are exactly zero will be counted in this bucket.

proto/openmetrics_data_model.proto
robskillington: Will update.

proto/openmetrics_data_model.proto
brian-brazil: But exclusive in the lower bound, thus it'll never have observations.

proto/openmetrics_data_model.proto
robskillington: This was removed and now reads `// The values must be monotonically increasing.`

robskillington:
@beorn7 fixed to allow Linear buckets to support negative observations as well as Explicit. Exponential still doesn't support negative observations but that's where we ended up - linear and explicit supporting it, but exponential not supporting it.

brian-brazil:
Line 138 needs updating for negative buckets too.

proto/openmetrics_data_model.proto
```diff
@@ -44,34 +46,43 @@ message MetricDescriptor {
...
+  // Required.
+  MetricType MetricType = 2;
+
   // STATE_SETs and INFO types must not have a unit.
   // Must match the regexp [a-zA-Z_:][a-zA-Z0-9_:]*
-  string unit = 2;
-
-  // Required.
-  Type type = 3;
+  string unit = 3;
```
brian-brazil, line 54: Must it be a suffix on the name?

proto/openmetrics_data_model.proto
```diff
@@ -238,6 +248,7 @@ message HistogramValue {
 
 // Value for STATE_SET point.
 message StateSetValue {
+  // Required, at least one.
```
brian-brazil, line 251: I don't think we ever said that you must have at least one (though it'd be weird not to).

proto/openmetrics_data_model.proto
```diff
@@ -44,34 +46,43 @@ message MetricDescriptor {
   // underscores are reserved for monitoring-system internal use.
   string name = 1;
 
+  // Required.
+  MetricType MetricType = 2;
+
   // STATE_SETs and INFO types must not have a unit.
   // Must match the regexp [a-zA-Z_:][a-zA-Z0-9_:]*
```
robskillington, line 53:
```suggestion
  // Must match the regexp [a-zA-Z_:][a-zA-Z0-9_:]*
  // If unit set, if MUST be a suffix on the metric name too (e.g. "foo_seconds").
```

proto/openmetrics_data_model.proto
robskillington, line 251:
```suggestion
  // Optional, should set at least one.
```

proto/openmetrics_data_model.proto
```diff
@@ -44,34 +46,44 @@ message MetricDescriptor {
...
-  string unit = 2;
-
-  // Required.
-  Type type = 3;
+  // If unit set, if MUST be a suffix on the metric name too (e.g. "foo_seconds").
```
robskillington, line 54:
```suggestion
  // If unit set, if MUST be a suffix on the metric name too. If the 
  // original name of the metric was "foo" and the unit was seconds 
  // then the name must be specified as "foo_seconds".
```

proto/openmetrics_data_model.proto
```diff
@@ -44,34 +46,46 @@ message MetricDescriptor {
...
+  MetricType MetricType = 2;
+
   // STATE_SETs and INFO types must not have a unit.
   // Must match the regexp [a-zA-Z_:][a-zA-Z0-9_:]*
-  string unit = 2;
-
-  // Required.
-  Type type = 3;
+  // If unit set, if MUST be a suffix on the metric name too. If the 
+  // original name of the metric was "foo" and the unit was seconds
```
brian-brazil, line 55: "seconds"

brian-brazil:
:+1:

proto/openmetrics_data_model.proto
robskillington, line 55:
```suggestion
  // original name of the metric was "foo" and the unit was "seconds" 
```

robskillington:
```suggestion
  // original name of the metric was "foo" and the unit was "seconds"  
```

PR #151, opened 2020-07-29
Update protobuf spec to include exemplar, encapsulate each metric type separately

robskillington:
cc @brian-brazil

### DIFF
--- a/proto/openmetrics_data_model.proto
+++ b/proto/openmetrics_data_model.proto
@@ -107,7 +107,7 @@ message Point {
   }
 
   // Optional, but SHOULD be specified for COUNTER, SUMMARY or
-  // CUMULATIVE_HISTOGRAM type.
+  // HISTOGRAM type.
   //
   // The cumulative value is over the time interval 
   // (start_timestamp, timestamp].
@@ -159,7 +159,7 @@ message CounterValue {
   Exemplar exemplar = 3;
 }
 
-// Value for CUMULATIVE_HISTOGRAM or GAUGE_HISTOGRAM point.
+// Value for HISTOGRAM or GAUGE_HISTOGRAM point.
 message HistogramValue {
   // Optional, but SHOULD be specified.
   oneof sum {
--- a/proto/openmetrics_data_model.proto
+++ b/proto/openmetrics_data_model.proto
@@ -50,11 +50,11 @@ message MetricFamily {
 
 // The type of a metric.
 enum MetricType {
-  // Unknown must use double point values.
+  // Unknown must use unknown point values, which is a double or int value.
   UNKNOWN = 0;
-  // Gauge must use double point values.
+  // Gauge must use double point values, which is a double or int value.
   GAUGE = 1;
-  // Counter must use counter point values, which is a double or int value.
+  // Counter must use counter point values, which is a double or uint value.
   COUNTER = 2;
   // State set must use state set value points.
   STATE_SET = 3;
@@ -97,12 +97,13 @@ message Label {
 message Point {
   // Required.
   oneof value {
-    double double_value = 1;
-    CounterValue counter_value = 2;
-    HistogramValue histogram_value = 3;
-    StateSetValue state_set_value = 4;
-    InfoValue info_value = 5;
-    SummaryValue summary_value = 6;
+    UnknownValue unknown_value = 1;
+    GaugeValue gauge_value = 2;
+    CounterValue counter_value = 3;
+    HistogramValue histogram_value = 4;
+    StateSetValue state_set_value = 5;
+    InfoValue info_value = 6;
+    SummaryValue summary_value = 7;
   }
 
   // Optional, but SHOULD be specified for COUNTER, SUMMARY or
@@ -113,11 +114,29 @@ message Point {
   // If start_timestamp is present and timestamp is absent,
   // the backend assigned timestamp may be used to construct the time
   // interval.
-  google.protobuf.Timestamp created = 7;
+  google.protobuf.Timestamp created = 8;
 
   // Optional.
   // If not specified, the timestamp will be decided by the backend.
-  google.protobuf.Timestamp timestamp = 8;
+  google.protobuf.Timestamp timestamp = 9;
+}
+
+// Value for UNKNOWN point.
+message UnknownValue {
+  // Required.
+  oneof value {
+    double double_value = 1;
+    int64 int_value = 2;
+  }
+}
+
+// Value for GAUGE point.
+message GaugeValue {
+  // Required.
+  oneof value {
+    double double_value = 1;
+    int64 int_value = 2;
+  }
 }
 
 // Value for COUNTER point.
--- a/proto/openmetrics_data_model.proto
+++ b/proto/openmetrics_data_model.proto
@@ -54,7 +54,7 @@ enum MetricType {
   UNKNOWN = 0;
   // Gauge must use double point values.
   GAUGE = 1;
-  // Counter must use double point values or int values.
+  // Counter must use counter point values, which is a double or int value.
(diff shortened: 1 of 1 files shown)


proto/openmetrics_data_model.proto
```diff
@@ -50,20 +50,20 @@ message MetricFamily {
 
 // The type of a metric.
 enum MetricType {
-  // Unknown must use double point values.
+  // Unknown must use unknown point values, which is a double or int value.
```
brian-brazil, line 53: We shouldn't have a mix of uint and int

proto/openmetrics_data_model.proto
```diff
@@ -50,20 +50,20 @@ message MetricFamily {
...
-  // Counter must use double point values or int values.
+  // Counter must use counter point values, which is a double or uint value.
   COUNTER = 2;
   // State set must use state set value points.
   STATE_SET = 3;
   // Info must use info value points.
   INFO = 4;
   // Cumulative histogram must use histogram value points.
-  CUMULATIVE_HISTOGRAM = 5;
+  HISTOGRAM_CUMULATIVE = 5;
```
brian-brazil, line 64: Just Histogram

proto/openmetrics_data_model.proto
```diff
@@ -50,20 +50,20 @@ message MetricFamily {
...
   // Gauge histogram must use histogram value points.
-  GAUGE_HISTOGRAM = 6;
+  HISTOGRAM_GAUGE = 6;
```
brian-brazil, line 66: Gauge histogram

proto/openmetrics_data_model.proto
```diff
@@ -113,11 +114,49 @@ message Point {
   // If start_timestamp is present and timestamp is absent,
   // the backend assigned timestamp may be used to construct the time
   // interval.
-  google.protobuf.Timestamp start_timestamp = 7;
+  google.protobuf.Timestamp created = 8;
```
brian-brazil, line 117: Why did this change tag number?

robskillington: This changed as I had to add explicit "UnknownValue", "GaugeValue" and "CounterValue" instead of the previous "double_value" and "int_value" (i.e. extra field in the `oneof` now).

brian-brazil: Tag values should never change in a proto, as that breaks backwards compatibility. Even though noone is likely using this proto, we should follow the rules.

robskillington:
Note: Need to also change sums to use oneof { double, int }.

proto/openmetrics_data_model.proto
robskillington: Since it's the first version it would be strange to have a missing tag number no? I would advocate for once we release it to not change the tag values, but up until that point making the proto as best as it can be rather than being concerned this is already in use somewhere somehow since it hasn't been published yet.

proto/openmetrics_data_model.proto
```diff
@@ -50,18 +50,18 @@ message MetricFamily {
 
 // The type of a metric.
 enum MetricType {
-  // Unknown must use double point values.
+  // Unknown must use unknown point values, which is a double or int value.
```
brian-brazil, line 53: There's no need for anything after the comma, that has its own definition.

proto/openmetrics_data_model.proto
```diff
@@ -50,18 +50,18 @@ message MetricFamily {
...
+  // Gauge must use gauge point values, which is a double or int value.
   GAUGE = 1;
-  // Counter must use double point values or int values.
+  // Counter must use counter point values, which is a double or uint value.
   COUNTER = 2;
   // State set must use state set value points.
   STATE_SET = 3;
   // Info must use info value points.
   INFO = 4;
   // Cumulative histogram must use histogram value points.
```
brian-brazil, line 63: Just Histogram.

proto/openmetrics_data_model.proto
```diff
@@ -97,12 +97,13 @@ message Label {
...
+    UnknownValue unknown_value = 1;
+    GaugeValue gauge_value = 2;
+    CounterValue counter_value = 3;
+    HistogramValue histogram_value = 4;
+    StateSetValue state_set_value = 5;
+    InfoValue info_value = 6;
+    SummaryValue summary_value = 7;
   }
 
   // Optional, but SHOULD be specified for COUNTER, SUMMARY or
```
brian-brazil, line 109: This needs updating.

proto/openmetrics_data_model.proto
```diff
@@ -113,17 +114,58 @@ message Point {
   // If start_timestamp is present and timestamp is absent,
   // the backend assigned timestamp may be used to construct the time
   // interval.
-  google.protobuf.Timestamp start_timestamp = 7;
+  google.protobuf.Timestamp created = 8;
```
brian-brazil, line 117: This doesn't apply to all types, this should be part of the relevant types only.

proto/openmetrics_data_model.proto
```diff
@@ -113,17 +114,58 @@ message Point {
...
+message GaugeValue {
+  // Required.
+  oneof value {
+    double double_value = 1;
+    int64 int_value = 2;
+  }
+}
+
+// Value for COUNTER point.
+// A Point in a Metric with the type Counter SHOULD have a Timestamp
```
brian-brazil, line 143: This and several other parts of this proto are duplicating the data model in the spec, and should be removed.

proto/openmetrics_data_model.proto
```diff
@@ -1,101 +1,49 @@
...
-  // Required.
   string name = 1;
 
-  // Required.
   string value = 2;
 }
 
-// A point in a metric.
 message Point {
-  // Required.
```
brian-brazil, line 98: We should probably leave all of these in, for the sake of anyone reading the proto alone.

proto/openmetrics_data_model.proto
```diff
@@ -106,246 +54,122 @@ message Point {
...
-  // the backend assigned timestamp may be used to construct the time
-  // interval.
   google.protobuf.Timestamp created = 3;
 
-  // Optional.
-  // An example of a value for a discrete event that was counted.
   Exemplar exemplar = 4;
 }
 
-// Value for HISTOGRAM or GAUGE_HISTOGRAM point.
```
brian-brazil: This one should probably stay, as it's the only types that share a value.

proto/openmetrics_data_model.proto
```diff
@@ -106,246 +54,122 @@ message Point {
...
-  // (created, timestamp].
-  // If created is present and timestamp is absent,
-  // the backend assigned timestamp may be used to construct the time
-  // interval.
   google.protobuf.Timestamp created = 3;
 
-  // Required.
   BucketOptions bucket_options = 4;
 
-  // `BucketOptions` describes the bucket boundaries used to create a
```
brian-brazil, line 134: All this bucket options stuff is specific to the proto, so if we still want all this then the comments should stay too.

proto/openmetrics_data_model.proto
```diff
@@ -175,81 +166,55 @@ message HistogramValue {
 
     // Specifies an exponential sequence of buckets that have an upper bound
     // that is proportional to the value of the upper bound of the previous
-    // bucket. This implies a constant relative uncertainty on the specific
```
brian-brazil, line 178: All this proto-specific stuff needs to stay, either that or get rid of Linear and Exponential as they only exist in the proto.

brian-brazil:
:+1:

robskillington:
@brian-brazil just confirming, this is good to merge?

brian-brazil:
Yes

PR #159, opened 2020-11-14
Fix Markdown syntax for Overview header

plant99:


plant99:
Found a few more typo errors, will remove the WIP tag when they're fixed.

dnsmichi (approved):
Thanks, came here for the same :)

RichiH:
Thanks!

PR #172, opened 2020-11-25
Fix Idnits 2.16.05 & 3.0.0 nits

RichiH:


RichiH:
Will split into easier-to-review commits

PR #195, opened 2021-04-09
Add URLs to tests

RichiH:


RichiH:
Self-merging so we can use it for Prometheus.

PR #201, opened 2021-06-08
Add makefile in /src

cw9:
Move golang related build targets to /src/Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ BUILD := $(abspath ./bin)
 .PHONY: test-impl
 test-impl:
 # openmetricstest needs to be built in /src since it requires /src/go.mod
-	(cd ./src && make openmetricstest)
+	cd ./src && make openmetricstest
 	$(BUILD)/openmetricstest -cmd-parser-text="$(cmd-parser-text)"
 
 .PHONY: prometheus_client_python_parser
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,16 @@
 BUILD := $(abspath ./bin)
-BINARIES :=         \
-	openmetricstest \
 
 # test-impl tests an OpenMetrics implementation.
 #
-# To test echo parser: 
+# To test echo parser:
 # make test-impl cmd-parser-text=echo
 #
 # To test github.com/prometheus/client_python parser:
 # make prometheus_client_python_parser test-impl cmd-parser-text="docker run --rm -i prometheus_client_python_parser:latest"
 .PHONY: test-impl
-test-impl: openmetricstest
+test-impl:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,16 +1,16 @@
...
+# To test echo parser:
 # make test-impl cmd-parser-text=echo
 #
 # To test github.com/prometheus/client_python parser:
 # make prometheus_client_python_parser test-impl cmd-parser-text="docker run --rm -i prometheus_client_python_parser:latest"
 .PHONY: test-impl
-test-impl: openmetricstest
+test-impl:
+# openmetricstest needs to be built in /src since it requires /src/go.mod
+	(cd ./src && make openmetricstest)
```
robskillington, line 13: nit: Think each line in the makefile executes as a subshell already so can do this without the brackets which is slightly cleaner. i.e. `cd ./src && make openmetricstest`

specification/OpenMetrics.txt
```diff
@@ -5,12 +5,12 @@
 Ops WG                                                  R. Hartmann, Ed.
 Internet-Draft                                              Grafana Labs
 Intended status: Standards Track                               B. Kochie
-Expires: June 20, 2021                                            GitLab
+Expires: December 10, 2021                                        GitLab
```
robskillington, line 8: Hm interesting, this date looks wrong. Maybe revert this file completely? It must have been regenerated during the move but I don't think we should regen it. This current version has been submitted for IETF for review.

src/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/cw9/OpenMetrics/src
```
robskillington, line 1: This should probably be `github.com/OpenObservability/OpenMetrics/src` before we merge I think?

If this is causing issues locally for you not matching your fork you can change the directory locally the repo is in and just add `github.com/cw9/OpenMetrics` as a git remote. Let me know if facing any issues with that.

robskillington:
Just minor comments, looks good 🙌

specification/OpenMetrics.txt
cw9: I have to regen it to pass the test

cw9: It was failing like this: https://app.circleci.com/pipelines/github/OpenObservability/OpenMetrics/119/workflows/b71981b0-1648-45e2-8e43-ecb5b57393d2/jobs/155

Makefile
cw9: done

src/go.mod
cw9: done

robskillington (approved):
LGTM

PR #206, opened 2021-06-14
Add scrape validator

cw9:
Added a scrape validator that validates metrics between scrapes.
Currently it validates that
- counter is not decreasing
- labels are not being duplicated across all metrics in a metric set.
- series are not disappearing
- timestamps not decreasing

Added a scrapeloop that runs in a loop to scrape and validate

TODOs in follow up diffs:
More validation rules.


src/cmd/scrapevalidator/scrape/scrape.go
```diff
@@ -0,0 +1,163 @@
...
+
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/pkg/textparse"
+	"github.com/prometheus/prometheus/pkg/timestamp"
+	"github.com/prometheus/prometheus/scrape"
+)
+
+type nowFn func() time.Time
+
+type ScrapeLoop struct {
```
robskillington, line 20: nit: We'll have to setup lint at some point, but just making a note we'll have to "stop the world" and go back and fix things like exported symbols without comments, etc at some point.

src/cmd/scrapevalidator/main.go
```diff
@@ -0,0 +1,23 @@
...
+	"flag"
+	"log"
+	"time"
+
+	"github.com/OpenObservability/OpenMetrics/src/cmd/scrapevalidator/scrape"
+)
+
+var (
+	endpoint       = flag.String("endpoint", "", "prom endpoint to validate")
+	scrapeTimeout  = flag.Duration("scrape-timeout", time.Second, "timeout for each scrape")
```
robskillington, line 13: Might want to up the scrape timeout to be at least half the scrape interval. Typically 80% or 75% of the scrape interval is used as the scrape timeout. (i.e. 8s).

src/cmd/scrapevalidator/scrape/scrape.go
```diff
@@ -0,0 +1,163 @@
...
+	validator      validator
+	scraper        scraper
+	scrapeTimeout  time.Duration
+	scrapeInterval time.Duration
+
+	nowFn nowFn
+}
+
+func NewScrapeLoop(
+	endpoint string,
```
robskillington, line 30: nit: Perhaps best to take `*url.URL` here and parse it in the command line main function? That way it's validated to be a valid URL before it starts to run rather than immediately returns an error.

When used can still use it by stringifying, i.e. `http.Get(url.String())`.

src/cmd/scrapevalidator/scrape/scrape.go
```diff
@@ -0,0 +1,163 @@
...
+type ScrapeLoop struct {
+	validator      validator
+	scraper        scraper
+	scrapeTimeout  time.Duration
+	scrapeInterval time.Duration
+
+	nowFn nowFn
+}
+
+func NewScrapeLoop(
```
robskillington, line 29: nit: Perhaps best to create a `ScrapeLoopOptions` struct to be passed here? I can see this needing to take a lot of different options over time and adjusting each callsite and every single unit test's constructor could get verbose with changes to the constructor args as they are introduced.

src/cmd/scrapevalidator/scrape/scrape.go
```diff
@@ -0,0 +1,163 @@
...
+			if err == io.EOF {
+				// Validate at the end of a scrape.
+				return numDataPointsFound, s.validator.Validate()
+			}
+			return 0, err
+		}
+		switch et {
+		case textparse.EntryType:
+			name, metricType := p.Type()
+			if err := tryResetMetadata(&dataPointFound, &m, yoloString(name)); err != nil {
```
robskillington, line 99: I think we should avoid all use of `yoloString` since it's an internal implementation detail whether that byte slice is longly lived or not (and therefore not clear if can safely keep the `[]byte` reference or not).

cw9: sg

src/cmd/scrapevalidator/scrape/scrape.go
cw9: sg, will do in a later diff for lint

src/cmd/scrapevalidator/scrape/scrape.go
cw9: I prefer to check in the first call because 1, not all errors can be found until the first call. 2, the first call is made almost immediately after the command runs.

src/cmd/scrapevalidator/scrape/scrape.go
cw9: sg

src/cmd/scrapevalidator/scrape/scrape.go
robskillington: SG

src/cmd/scrapevalidator/scrape/validator.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	last, ok := v.curMetricSet[key]
+	if !ok {
+		v.curMetricSet[key] = cur
+		return nil
+	}
+	if err := validate(last, cur); err != nil {
+		return err
+	}
+	return nil
```
robskillington, line 78: nit: This can just be `return validate(last, cur)`?

robskillington (approved):
LGTM

PR #207, opened 2021-06-16
Differentiate error level

cw9:


src/cmd/scrapevalidator/main.go
```diff
@@ -10,23 +10,33 @@ import (
 )
 
 var (
-	endpoint       = flag.String("endpoint", "", "prom endpoint to validate")
-	scrapeTimeout  = flag.Duration("scrape-timeout", 8*time.Second, "timeout for each scrape")
-	scrapeInterval = flag.Duration("scrape-interval", 10*time.Second, "time between scrapes")
+	endpointArg       = flag.String("endpoint", "", "prom endpoint to validate")
```
robskillington, line 13: nit: Perhaps put `(required)` at the end of the description help text (given that without it the usage is printed and it exits)?

robskillington (approved):
LGTM

PR #210, opened 2021-06-16
run test cases with the custom validator

cw9:
Move the validator in the scrape loop out to its own pkg
Create a tool to validate input based on the validator

created build target
`make test_open_metrics_validator` which runs the existing test cases against the new tool, which behaves better than the   native golang parser we used in `make test_prometheus_client_go_parser`, the validator fails 71 test cases while the native golang parser fails 93

We will continue to improve on the validator to fix all the tests


robskillington (approved):
LGTM

PR #211, opened 2021-06-16
More validation

cw9:


src/validator/validator.go
```diff
@@ -195,42 +231,68 @@ func tryResetMetadata(dataPointFound *bool, m *scrape.MetricMetadata, name strin
...
-	last, ok := v.curMetricSet[key]
+	key := labelKey(lset)
+	last, ok := mf.metrics[key]
 	if !ok {
-		v.curMetricSet[key] = cur
+		mf.metrics[key] = cur
 		return nil
 	}
-	return validate(last, cur)
+	return compareMetric(*mf.metricType, last, cur)
```
robskillington, line 270: Do you need to check if `mf.MetricType` is nil or not before de-referencing it with `*`? 

By reading the code it seems plausible that you could be processing a metric that hasn't had it's type set on the metric family?

src/validator/validator.go
```diff
@@ -266,23 +330,94 @@ func duplicatedLabels(this, other labels.Labels) labels.Labels {
...
 
-// validate validates the current record against last record for a metric.
-// TODO: validate more metric types.
-func validate(last, cur metricDataPoint) error {
-	switch last.m.Type {
+func (v *OpenMetricsValidator) compareMetricFamilies(last, cur *metricFamily) error {
+	for lset, lastMF := range last.metrics {
+		curMF, ok := cur.metrics[lset]
+		if ok {
+			return compareMetric(*cur.metricType, lastMF, curMF)
```
robskillington, line 337: Here as well, do you need to check for nil before de-referencing? Maybe `compareMetric` could even just take a pointer type so that you can do the nil check from within the `compareMetric` method and share the logic instead of implementing with each caller before it calls `compareMetric`?

src/validator/validator.go
```diff
@@ -266,23 +330,94 @@ func duplicatedLabels(this, other labels.Labels) labels.Labels {
...
+		val := m.lset.Get(labels.BucketLabel)
+		if val == "+Inf" {
+			positiveInfBucketFound = true
+		}
+	}
+	if !positiveInfBucketFound {
+		return errMustContainPositiveInfBucket
+	}
+	return nil
+}
```
robskillington, line 402: Do these need to do the `myErr.tryReport(...)` logic?

src/validator/validator.go
```diff
@@ -266,23 +330,94 @@ func duplicatedLabels(this, other labels.Labels) labels.Labels {
...
+
+// compareMetric validates the current record against last record for a metric.
+// TODO: compareMetric more metric types.
+func compareMetric(mt textparse.MetricType, last, cur metric) error {
+	if cur.timestamp <= last.timestamp {
+		return errMustTimestampIncrease
```
robskillington, line 408: Do this need to do the `myErr.tryReport(...)` logic too?

src/validator/validator.go
cw9: I can check it, but it should an invalid metric family if the type is missing yea?

robskillington (approved):
LGTM other than minor comments

src/validator/validator.go
cw9: I skipped the tryReport() check for `must` errors, only been calling tryReport on `should` errors

src/validator/validator.go
robskillington: Correct yeah, just unsure if you want it to panic or return an error. If you want it to not panic I assume we'll have to check if nil or not?

src/validator/validator.go
robskillington: Np, sounds good

src/validator/validator.go
```diff
@@ -46,10 +46,16 @@ var (
 		level: ErrorLevelMust,
 	}
 
+	errMustMetricFamilyWithMetadata = errorWithLevel{
+		err:   errors.New("A MetricFamily MUST have a name, HELP, TYPE, and UNIT metadata"),
```
robskillington, line 50: Technically I think it only needs a TYPE yeah?

robskillington:
Looks good 👍

PR #213, opened 2021-06-21
Capture all errors

cw9:


src/cmd/scrapevalidator/scrape/loop.go
```diff
@@ -55,36 +55,38 @@ func NewLoop(
...
 
-func (l *Loop) runOnce() error {
+func (l *Loop) runOnce() {
 	ctx, cancel := context.WithTimeout(context.Background(), l.scrapeTimeout)
 	defer cancel()
 
 	b, err := l.scraper.Scrape(ctx)
 	if err != nil {
-		return err
+		log.Printf("scrape failed: %v", err)
```
robskillington, line 81: Probably needs an endline after `%v` since using Printf (and not println)?

src/cmd/scrapevalidator/scrape/loop.go
```diff
@@ -55,36 +55,38 @@ func NewLoop(
...
+		return
 	}
 	log.Println("scraped successfully")
 
-	err = l.validator.Validate(b)
-	if err == nil {
-		log.Println("validated successfully")
+	if err := l.validator.Validate(b); err != nil {
+		l.validator.Reset()
+		log.Printf("validation failed: %v", err)
```
robskillington, line 88: Probably needs an endline after `%v` since using Printf (and not println)?

src/validator/validator.go
```diff
@@ -336,42 +328,38 @@ func duplicatedLabels(this, other labels.Labels) labels.Labels {
...
 	case textparse.MetricTypeHistogram, textparse.MetricTypeGaugeHistogram:
-		return v.validateMetricFamilyHistogram(cur)
+		v.validateMetricFamilyHistogram(mn, cur)
 	case textparse.MetricTypeSummary:
-		return v.validateMetricFamilySummary(cur)
+		v.validateMetricFamilySummary(mn, cur)
 	case textparse.MetricTypeStateset:
-		return v.validateMetricFamilyStateSet(cur)
+		v.validateMetricFamilyStateSet(mn, cur)
 	default:
```
robskillington, line 350: nit: default case no longer required?

src/validator/validator.go
```diff
@@ -402,32 +390,36 @@ func (v *OpenMetricsValidator) validateMetricFamilyHistogram(cur *metricFamily)
...
+		v.addError(mn, errMustNotCounterValueDecrease)
 	}
-	return nil
+}
+
+func (v *OpenMetricsValidator) addError(mn string, err error) {
+	if err != nil {
+		err = fmt.Errorf("error found on metric %q: %v", mn, err)
+	}
+	v.mErr = multierr.Append(v.mErr, err)
```
robskillington, line 422: Does this take into account if err == nil?

Could also just return early from method if err is nil and only append if not nil?

robskillington (approved):
LGTM other minor comments

src/validator/validator.go
cw9: yep

src/validator/validator.go
cw9: it handles err == nil properly too, but probably the way you suggested is more clear, will fix

PR #218, opened 2021-06-29
Pull in latest timestamp parsing fix in OpenMetrics Parser

cw9:


tests/testdata/parsers/duplicate_timestamps/metrics
```diff
@@ -1,8 +1,8 @@
...
-a{a="1",foo="bar"} 1 0.0000000000
-a{a="1",foo="bar"} 2 0.0000000001
-a{a="1",foo="bar"} 3 0.0000000010
-a{a="2",foo="bar"} 4 0.0000000000
-a{a="2",foo="bar"} 5 0.0000000001
+a{a="1",foo="bar"} 1 0
+a{a="1",foo="bar"} 2 0
+a{a="1",foo="bar"} 3 0
+a{a="2",foo="bar"} 4 0
+a{a="2",foo="bar"} 5 0
```
cw9, line 7: This is not an error from the python test btw, but is not allowed in the spec: https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric

robskillington (approved):
LGTM

PR #220, opened 2021-07-01
Allow non-decreasing timestamps

cw9:


tests/testdata/parsers/duplicate_timestamps/test.json
```diff
@@ -1,5 +1,5 @@
 {
   "type": "text",
   "file": "metrics",
-  "shouldParse": false
+  "shouldParse": true
```
robskillington, line 4: Should we also restore the test scrape payload? Or maybe we should make two of them, i.e. `duplicate_timestamps_0`, `duplicate_timestamps_1`?

cw9: done

tests/testdata/parsers/bad_exemplar_timestamp_Inf/metrics
```diff
@@ -0,0 +1,3 @@
+# TYPE a counter
```
robskillington, line 1: Maybe we call this path `bad_exemplar_timestamp_0` to following naming of other test cases (which do not have the specific name of the test in the path?

I agree it is probably better, but better perhaps to stay consistent and either keep the old naming of `bad_category_$N` and then in a followup PR change it to a new way across everything (later)?

robskillington (approved):
LGTM

PR #221, opened 2021-07-01
Add testing of openmetricsvalidator and clientpythonparser in CI and remove raw go parser

cw9:


specification/OpenMetrics.txt
```diff
@@ -1107,7 +1107,7 @@ Internet-Draft                 OpenMetrics                     July 2021
    false.
 
    An example with the states "a", "bb", and "ccc" in which only the
-   value b is enabled and the metric name is foo:
+   value bb is enabled and the metric name is foo:
```
robskillington, line 1110: Good catch.

robskillington (approved):
LGTM, great that we're testing the implementations now too

PR #222, opened 2021-07-14
Add definition of IETF

ChrisChinchilla:
I was reading through this page and came across IETF and had no idea what it was, so expanded the acronym in the first use. Small, but hopefully helpful.

### DIFF
--- a/specification/OpenMetrics.md
+++ b/specification/OpenMetrics.md
@@ -48,7 +48,7 @@ informative:
 
 --- abstract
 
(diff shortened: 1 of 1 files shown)


RichiH:
Thanks, makes sense. Can you flip it around, please? Acronym first, expansion in parentheses.

ChrisChinchilla:
@RichiH Most style guides (e.g. https://docs.microsoft.com/en-us/style-guide/acronyms) recommend this way around, but I can if you're set on that way around :)

RichiH:
WFM; I learned it the other way around, but it does not really matter either way - and it's provably better.

Thanks!

Issue #223, opened 2021-08-05
Inconsistent definition of `_created` timestamp for GaugeHistogram


fstab:
Under [#suffixes](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#suffixes) there's a list of suffixes for metric types:

* Counter: '_total', '_created'
* Summary: '_count', '_sum', '_created', '' (empty)
* Histogram: '_count', '_sum', '_bucket', '_created'
* GaugeHistogram: '_gcount', '_gsum', '_bucket'
* Info: '_info'
* Gauge: '' (empty)
* StateSet: '' (empty)
* Unknown: '' (empty)

This looks like `GaugeHistogram` does not have a `_created` timestamp.

However, if you look at [#gaugehistogram](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#gaugehistogram-1) the `_created` value is explicitly mentioned in the description and shown in the example. So, can `GaugeHistogram` have a `_created` timestamp?

This makes me wonder about `_created` for other types as well. Can you conclude from the OpenMetrics specification that `Gauge`, `StateSet`, `Info`, and `Unknown` must not have a `_created` timestamp, or is there no such restriction?

brian-brazil:
> This looks like GaugeHistogram does not have a _created timestamp.

Correct, only counter-like metrics can have them.

> However, if you look at #gaugehistogram the _created value is explicitly mentioned in the description and shown in the example.

That is a mistake, which should be fixed.

brian-brazil:
Fixed in #226

PR #226, opened 2021-09-22
List '_created' under GaugeHistogram suffixes

vpranckaitis:
Further down the  document, `_created` is mentioned as a possible suffix for GaugeHistograms

https://github.com/OpenObservability/OpenMetrics/blob/45ae6699a4ba83368f0b481f05e798ba3c398eec/specification/OpenMetrics.md#gaugehistogram-1

https://github.com/OpenObservability/OpenMetrics/blame/45ae6699a4ba83368f0b481f05e798ba3c398eec/specification/OpenMetrics.md#L793


vpranckaitis:
Seems that `OpenMetrics.{md,txt}` diverged, not sure how to deal with that.

brian-brazil:
This is incorrect per #223, GaugeHistograms cannot have a _created as they are not counter-like. Would you like to fix it the other way around?

brian-brazil:
Thanks!

Issue #229, opened 2021-09-24
Tests expect histogram behaviour not outlined in the spec


sinkingpoint:
# _count can't exist without _sum (and visa versa)

In the [Histogram Section](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#histogram) the spec says:

```
A Histogram MetricPoint MUST contain at least one bucket, and SHOULD contain Sum, and Created values
```

but then we have tests like [bad_histograms_1](https://github.com/OpenObservability/OpenMetrics/blob/main/tests/testdata/parsers/bad_histograms_1/metrics) that have a _sum without a _count (Note that both are "SHOULD contain" not "MUST contain". The Python implementation in this repo says that this errors because a histogram cannot have a _sum without a _count, but this doesn't seem to be formalised anywhere. In addition, that requirement contradicts the statement that:

```
Negative threshold buckets MAY be used, but then the Histogram MetricPoint MUST NOT contain a sum value as it would no longer be a counter semantically
```
i.e. is a _count without a _sum (because we have negative buckets) valid?

# Can't have a negative _sum without negative buckets

Additionally, in the [bad_histograms_7](https://github.com/OpenObservability/OpenMetrics/blob/main/tests/testdata/parsers/bad_histograms_7/metrics) test, we have a negative sum with the python implementation saying that sum cannot be negative without negative buckets but this restriction is not mentioned in https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#gaugehistogram (however that might be implied by the definition of _sum, IMO it should still be called out)

# Histograms can't have negative sums with negative buckets?

Lastly, to me it's kind of confusing that GaugeHistograms _can_ have negative _sums and Histograms can't, even though both can have negative buckets. In the Histograms section, the spec says

```
Semantically, Sum, and buckets values are counters
```

Should that be _Count_ and bucket values are counters?

If you consider a histogram like:

```
# TYPE a histogram
# HELP a A test histogram
a_bucket{le="-3"} 1
a_bucket{le="+Inf"} 1
```

then my one observed value might have been -5, which means my `a_count` would be `1` and my `a_sum` would be -5. This makes sense to me (but maybe I'm misinterpreting), and doesn't violate the semantics of _sum IMO

sinkingpoint:
This got a bit long - I could split it out into separate issues if that would be easier

brian-brazil:
[...]

PR #232, opened 2021-09-24
Fix bad_grouping_or_ordering_0 test

sinkingpoint:
This test is really janky in that it actually tests 4 things,
rather than just the metric interweaving it proports to test:

- "A Histogram MetricPoint MUST contain at least one bucket (the given one doesn't)
- "Histogram MetricPoints MUST have at least a bucket with an +Inf threshold" (the given one doesn't)
- The sum without a count thing I mentioned in #229 (the histograms have sums without counts)
- The metric interweaving (what it's meant to be testing)

This commit fixes the first three to make this only test one thing


brian-brazil:
That would make sense, the test was designed to test the specific code of the Python parser. Can you fix the Go validator so it correctly rejects this?

brian-brazil:
I think you forgot to push the changes to the Go code itself.

sinkingpoint:
Sure. Think that commit should do it. It's not the cleanest, but it seems to work

src/validator/validator.go
```diff
@@ -535,6 +552,24 @@ func (v *OpenMetricsValidator) compareMetricFamilies(mfn string, last, cur *metr
...
+func getIgnoredLabels(name string, cur *metricFamily) []string {
+	ignored := []string{}
+	ignored = append(ignored, "__name__")
+	switch cur.MetricType() {
+	case textparse.MetricTypeHistogram:
+	case textparse.MetricTypeGaugeHistogram:
+		if strings.HasSuffix(name, "_bucket") {
+			ignored = append(ignored, labels.BucketLabel)
+		}
+	case textparse.MetricTypeStateset:
```
brian-brazil, line 564: I think you meant summary. Also there's no suffix on a quantile, you want to check the sample metric name matches the metric family name.

src/validator/validator.go
```diff
@@ -535,6 +552,24 @@ func (v *OpenMetricsValidator) compareMetricFamilies(mfn string, last, cur *metr
 	}
 }
 
+func getIgnoredLabels(name string, cur *metricFamily) []string {
+	ignored := []string{}
+	ignored = append(ignored, "__name__")
+	switch cur.MetricType() {
+	case textparse.MetricTypeHistogram:
```
brian-brazil, line 559: Fallthroughs must be explicit in Go

sinkingpoint:
Fixed those. Thanks for all your guidance @brian-brazil

brian-brazil:
Thanks!

PR #233, opened 2021-10-21
docs: Clarify number of +Inf buckets in *histogram metric types

millermatt:
The text "MUST have one bucket with an +Inf threshold" is easier to comprehend than the text "MUST have at least a bucket with an +Inf threshold" while communicating the same information.


brian-brazil:
Thanks!

PR #239, opened 2022-02-16
Fix test cases that were failing due to bad bucket label instead of intended failures

robskillington:
Also does the following:
- Fixes the discovered gap of exemplar length not being checked by the Go validator (the Python client parser correctly checks length).
- Fixes more discovered gaps for Go validator such as validation of bucket order and gauge histogram bucket/gsum values, counter-like values non-nan and non-negative, etc.
- Add the ability to add comments to the test cases in the test case definition.


SuperQ:
Whups, it looks like it's not being caught.

robskillington:
Brian pointed out that actually we had a test case for it (bad_exemplar_6) however there was a bad bucket label value (of `" Inf"` instead of `"+Inf"`) which was also causing the test to fail and masking the gap for the Go validator @cw9 had put together (the Python client library was actually catching this however, but the test wasn't providing coverage since the test was also failing due to the bad bucket label value).

tests/testdata/parsers/bad_histograms_9/bad_histograms_5/metrics.txt
```diff
@@ -1,4 +0,0 @@
-# TYPE a gaugehistogram
```
brian-brazil, line 1: Why is this being deleted? There's a gcount without a gsum.

tests/testdata/parsers/bad_histograms_10/bad_histograms_6/metrics.txt
```diff
@@ -1,4 +0,0 @@
-# TYPE a histogram
-a_count 1
-a_bucket{le=" Inf"} 0
```
brian-brazil, line 3: Why is this being deleted? The count doesn't match the +Inf

src/validator/validator.go
```diff
@@ -742,13 +760,33 @@ func (v *OpenMetricsValidator) validateMetric(mn string, mt textparse.MetricType
...
 	switch mt {
+	case textparse.MetricTypeCounter:
+		v.validateMetricCounterValue(mn, cur)
+	case textparse.MetricTypeHistogram:
+		if strings.HasSuffix(mn, "_count") || strings.HasSuffix(mn, "_sum") || strings.HasSuffix(mn, "_created") || strings.HasSuffix(mn, "_bucket") {
+			v.validateMetricCounterValue(mn, cur)
+		}
+	case textparse.MetricTypeGaugeHistogram:
+		if strings.HasSuffix(mn, "_gcount") || strings.HasSuffix(mn, "_gsum") || strings.HasSuffix(mn, "_bucket") {
+			v.validateMetricCounterValue(mn, cur)
```
brian-brazil, line 771: In gauge histograms, none of these are counters. What you're checking for here is NaN, not whether it's a counter.

src/validator/validator.go
```diff
@@ -742,13 +760,33 @@ func (v *OpenMetricsValidator) validateMetric(mn string, mt textparse.MetricType
...
+		}
+	case textparse.MetricTypeSummary:
+		if strings.HasSuffix(mn, "_count") || strings.HasSuffix(mn, "_sum") || strings.HasSuffix(mn, "_created") {
```
brian-brazil, line 774: Created is not a counter.

src/validator/validator.go
```diff
@@ -742,13 +760,33 @@ func (v *OpenMetricsValidator) validateMetric(mn string, mt textparse.MetricType
 		}
 	}
 	switch mt {
+	case textparse.MetricTypeCounter:
+		v.validateMetricCounterValue(mn, cur)
```
brian-brazil, line 764: Created is not a counter.

tests/testdata/parsers/bad_histograms_9/bad_histograms_5/metrics.txt
robskillington: This was a duplicate test (note the path which shows the unexpected nesting `tests/testdata/parsers/bad_histograms_9/bad_histograms_5/metrics.txt`).

It still exists here:
https://github.com/OpenObservability/OpenMetrics/blob/1ca3b0f10dd8fe7fd1226302fb794ce2a25e47bd/tests/testdata/parsers/bad_histograms_5/metrics#L1-L4

tests/testdata/parsers/bad_histograms_10/bad_histograms_6/metrics.txt
robskillington: Similary, this was a duplicate test. Still exists here:
https://github.com/OpenObservability/OpenMetrics/blob/1ca3b0f10dd8fe7fd1226302fb794ce2a25e47bd/tests/testdata/parsers/bad_histograms_10/metrics#L1-L5

src/validator/validator.go
robskillington: TY sg, corrected.

src/validator/validator.go
robskillington: Did not think this one through, thanks 👍

src/validator/validator.go
robskillington: Did not think this one through, thanks 👍

tests/testdata/parsers/bad_histograms_9/bad_histograms_5/metrics.txt
brian-brazil: That looks like a different test

tests/testdata/parsers/bad_histograms_10/bad_histograms_6/metrics.txt
brian-brazil: That's a different test

tests/testdata/parsers/bad_histograms_9/bad_histograms_5/metrics.txt
robskillington: Hm, true. I'm going to have to move this test though, it needs to be renamed since it clashes with the one I linked to (they were both called `bad_histograms_5`).

tests/testdata/parsers/bad_histograms_10/bad_histograms_6/metrics.txt
robskillington: Ok yeah, renamed, fixed the " Inf" and fixed the directory structure - now this is:
https://github.com/OpenObservability/OpenMetrics/blob/13530e4570d42d2e2d2a75d8803992c83447cba1/tests/testdata/parsers/bad_histograms_13/metrics

tests/testdata/parsers/bad_histograms_9/bad_histograms_5/metrics.txt
robskillington: Ok so renamed, fixed the " Inf" and fixed the directory structure, this file is now here:
https://github.com/OpenObservability/OpenMetrics/blob/13530e4570d42d2e2d2a75d8803992c83447cba1/tests/testdata/parsers/bad_histograms_14/metrics

PR #240, opened 2022-02-18
Clarify a few things around the spec

brian-brazil:
Based on various feedback over the last year clarify some wording, and where spec and code deviated.
See commits for more details.


specification/OpenMetrics.md
```diff
@@ -440,6 +441,9 @@ Line feed, '\n' (0x0A) -> literally '\n' (Bytecode 0x5c 0x6e)
 Double quotes -> '\"' (Bytecode 0x5c 0x22)
 Backslash -> '\\' (Bytecode 0x5c 0x5c)
 
+You SHOULD NOT use a backslash before other characters, as they do not escape.
+'\\a' SHOULD be used instead of the equivialent '\a' for example.
+
```
RichiH, line 446:
```suggestion
Double backslashes SHOULD be used to escape characters. Single backslash SHOULD NOT be used, but will work for well-defined escape sequences.
As an example, '\\a' escapes and is preferable to '\a'. As another example, '\\n' is preferable even though '\n' also escapes.

```

RichiH (approved):
LGTM other than suggested change.

specification/OpenMetrics.md
brian-brazil: Your suggestion doesn't make sense as-is, but I've taken the thrust of it.

specification/OpenMetrics.md
```diff
@@ -440,6 +441,10 @@ Line feed, '\n' (0x0A) -> literally '\n' (Bytecode 0x5c 0x6e)
 Double quotes -> '\"' (Bytecode 0x5c 0x22)
 Backslash -> '\\' (Bytecode 0x5c 0x5c)
 
+A double backslashe SHOULD be used to represent a backslash character.
+Single backslash SHOULD NOT be used for undefined escape sequences, but will work.
+As an example, '\\a' is equivilent and preferable to '\a'.
```
RichiH, line 446:
```suggestion
A double backslash SHOULD be used to represent a backslash character.
A single backslash SHOULD NOT be used for undefined escape sequences.
As an example, '\\a' is equivalent and preferable to '\a'.
```

brian-brazil: Fixed, and also spotted escaping issues in rendered version which I've fixed too.

RichiH (approved):

PR #254, opened 2022-09-15
Remove not true statement about OpenMetrics being the official standard in Prometheus

bogdandrutu:
* Not all Prometheus clients support exporting OpenMetrics, and not all Prometheus clients do this by default.
* Prometheus Text 0.4 is not compatible with OpenMetrics for things like Meta for Counters, etc.

Because of these things I think OpenMetrics should remove this claim, until Prometheus properly adopt OpenMetrics.

Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>

### DIFF
--- a/website/config.yaml
+++ b/website/config.yaml
@@ -5,7 +5,7 @@ disableKinds: [taxonomy, taxonomyTerm]
 
 params:
   description: |
-    OpenMetrics is the official standard for metrics in the [Prometheus](https://prometheus.io/) ecosystem. With its support for examplars, it ties metrics and traces as closely and efficiently as never seen before. The [specification](https://github.com/OpenObservability/OpenMetrics/blob/master/specification/OpenMetrics.md) covers semantics, data model, and wire format for cloud-native metrics at scale, with support for both text representation and Protocol Buffers.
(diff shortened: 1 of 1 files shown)


jeromeinsf:
@RichiH ?

mxinden:
Adding a data point here, the new [Rust Prometheus client](https://github.com/prometheus/client_rust) supports **only** the Open Metrics format.

RichiH:
This is the wrong repository; it's for Prometheus to decide what the Prometheus project wants, not for OpenMetrics.

@bogdandrutu would you like to put this on the backlog for the Prometheus Dev Summits? Will you make it to Munich in November?

bogdandrutu:
@RichiH sorry to say, but this is not a good answer to close this issue, I know you guys are unarguably the best in understanding all of these, but throwing the problem on Prometheus is not a solution. This repo makes a claim, is not Prometheus problem that you make a claim

bogdandrutu:
Also, @RichiH you did add this text https://github.com/OpenObservability/OpenMetrics/commit/f00a4920acb02fec2db45d7f709ed44da162175e so I would like to understand what was the rational to add such a strong text.

PR #256, opened 2022-10-05
Native histogram support in proto spec

beorn7:
This is marked as WIP because it is not meant as a final proposal yet, merely to demonstrate how it could look like and enable early adopters to play with it and for the reviewers to give feedback.

I am not sure what the stability guarantees are for the proto spec. I believe that these changes should not be a breaking change WRT the wire format, but they do change the API, mostly because of the different naming of `oneof` fields and turning the count field of a histogram into a `oneof` (which would have created a naming collision if I didn't change the naming).

In different news, the previous naming of the `oneof` fields seems like an oversight to me (the author might not have been aware of the resulting naming of the getters, where the top-level name of the `oneof` block doesn't show up at all, e.g. in Go, the getter for the histogram `sum` field was called `GetDoubleValue` previously).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ test_open_metrics_validator:
 
 .PHONY: proto_go
 proto_go: setup
-	protoc --go_out=$(BUILD) --go_opt=paths=source_relative ./proto/*.proto
+	protoc --go_out=. --go_opt=paths=source_relative ./proto/*.proto
 
 .PHONY: setup
 setup:
--- a/proto/openmetrics_data_model.proto
+++ b/proto/openmetrics_data_model.proto
@@ -110,8 +110,8 @@ message GaugeValue {
 message CounterValue {
   // Required.
   oneof total {
-    double double_value = 1;
-    uint64 int_value = 2;
+    double double_total = 1;
+    uint64 int_total = 2;
   }
 
   // The time values began being collected for this counter.
@@ -126,8 +126,8 @@ message CounterValue {
 message HistogramValue {
   // Optional.
   oneof sum {
-    double double_value = 1;
-    int64 int_value = 2;
+    double double_sum = 1;
+    int64 int_sum = 2;
   }
 
   // Optional.
@@ -190,8 +190,8 @@ message InfoValue {
 message SummaryValue {
   // Optional. 
   oneof sum {
-    double double_value = 1;
-    int64 int_value = 2;
+    double double_sum = 1;
+    int64 int_sum = 2;
   }
 
   // Optional. 
--- a/proto/openmetrics_data_model.proto
+++ b/proto/openmetrics_data_model.proto
@@ -5,6 +5,8 @@ syntax = "proto3";
 // All string fields MUST be UTF-8 encoded strings.
 package openmetrics;
 
+option go_package = "github.com/OpenObservability/OpenMetrics/proto;openmetrics";
+
 import "google/protobuf/timestamp.proto";
 
 // The top-level container type that is encoded and sent over the wire.
--- a/proto/openmetrics_data_model.proto
+++ b/proto/openmetrics_data_model.proto
@@ -131,7 +131,10 @@ message HistogramValue {
   }
 
   // Optional.
-  uint64 count = 3;
+  oneof count {
+    double double_count = 6;
+    uint64 int_count = 3;
+  }
 
   // The time values began being collected for this histogram.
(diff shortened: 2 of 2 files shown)


brian-brazil:
From a very quick glance this looks okay for a future OM 2.0, other than the `schema` being required as the old way should still work. It probably only needs a little verbiage in the comments on how to tell which way is in use.

We may wish to do the renamings of the existing fields now in 1.x, to avoid future hassle when things are harder to change. I'm okay with that in principle as it should be fairly low impact now, however it may count as a breaking change. If the current Go accessors aren't right, we should probably fix that in 1.x though.

beorn7:
Right, `schema` has to be optional so that you can do it the old way. Will update…

bwplotka:
What's the latest on this? It would be awesome to have native histogram support in proto 💪🏽

beorn7:
I believe this has everything native histograms need. This seems also be generally correct from the OM side. At least, after the first correction, no other concerns have been voiced. I have marked it as "ready for review" now. However

However, this needs some shepherding from the OM side:
1. Final correctness review from the OM side.
2. What version bumps of the standard are required to merge this, if any?
3. How to get this in sync with the text format? And is that needed before merging this?

The 3rd item refers to the efforts to get native histograms into the OM text format. Paging @csmarchbanks and @SuperQ .

beorn7:
@SuperQ @RichiH could you advise how to move this forward?

The v1.0.0 protobuf spec is part of https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md , so we could even merge this without changing v1.0.0. But maybe you would like a v1.1 or v2.0 branch or something?

Note, however, that the current proto file has naming issues that you might want to fix within v1.0.0 or as v1.0.1 (it's only a change in the generated code, not in the wire format of the protobuf, see above: "… the previous naming of the oneof fields seems like an oversight to me (the author might not have been aware of the resulting naming of the getters, where the top-level name of the oneof block doesn't show up at all, e.g. in Go, the getter for the histogram sum field was called GetDoubleValue previously)."

RichiH:
Proto looks fine to me; I'd prefer to have the same changes in the textual description as well though. Is there a short-form prose which I could adopt into normative language?

beorn7:
The best right now is probably the comments in the proto file (which are also contained in this PR).

[This](https://github.com/prometheus/docs/blob/beorn7/histogram/content/docs/concepts/native_histograms.md) is some kind of histogram spec, that I'm currently working on. You could pick from the data model (which is mostly complete) but the part specifically about the exposition format is still work in progress.

kakkoyun (approved):

beorn7:
Given that we do not aim for a separate OM proto spec anymore, I'm closing this.

costela:
@beorn7 what does this mean for the future of e.g. native histograms in the OTEL ecosystem? AFAICT on the OTEL side [there hasn't been any updates](https://github.com/open-telemetry/opentelemetry-specification/blob/704bee62bf3f5921b7bde4875c91df7bd9e8920b/specification/compatibility/prometheus_and_openmetrics.md?plain=1#L76-L77) and the expectation was that the spec changes would be driven here?

Thanks for the update :pray: 


UPDATE: fixed link :see_no_evil:

beorn7:
The Prometheus protobuf based exposition format already supports native histograms and all other features. It will be _the_ canonical protobuf based exposition format.

It's essentially just a rebranding: OpenMetrics implies text-based format. Previously, we planned to also have a protobuf-based format under the OpenMetrics brand.

See https://github.com/prometheus/OpenMetrics/issues/296 for details.

Having said that, I don't understand why this should affect "native histograms in the OTel ecosystem". Technically, there are no "native histograms" in OTel. There are "exponential histograms", which are 100% compatible with Prometheus "native histograms" with a standard exponential schema.

Furthermore, you have linked to the parquet-go repository, where I don't get the connection, either.

costela:
@beorn7 oh boy, sorry, copy-paste f*ck-up. I meant [this link](https://github.com/open-telemetry/opentelemetry-specification/blob/704bee62bf3f5921b7bde4875c91df7bd9e8920b/specification/compatibility/prometheus_and_openmetrics.md?plain=1#L76-L77)

I was probably mislead by "OpenMetrics == text-base" + "(Native) Histograms are not currently supported in the Prometheus text exposition format" (from the link above).

So there's nothing to see here; sorry for the noise! :pray:

PR #257, opened 2022-11-03
Update project status (changed on 3rd Feb 2022)

SphinxKnight:
cf https://www.cncf.io/blog/2022/02/03/cncf-cultivated-openmetrics-becomes-an-incubating-project/


brian-brazil:
Good spot, thanks for fixing!

PR #267, opened 2024-02-26
Allow Exemplars on all time series

fstab:
At the [Prometheus dev summit on 2022-11-10](https://docs.google.com/document/d/11LC3wJcVk00l8w5P3oLQ-m3Y37iom6INAMEu2ZAGIIE/edit#heading=h.lwq7fdwwtszy) there was consensus that

>  Prometheus will ingest Exemplars on all time series.

Previously exemplars were only allowed for counters and histogram buckets.

The change has been implemented in the Prometheus server and was merged on Mar 5, 2023 with prometheus/prometheus#11984.

This PR updates the OpenMetrics spec to explicity allow exemplars on all time series.

This is not a breaking change for client libraries producing OpenMetrics text format. Exemplars are optional, and time series without exemplars are of course still valid.

It might however break scrapers parsing OpenMetrics if these scrapers verify that exemplars can only be present for counters and histogram buckets.

Because of the potential breaking change, we create a new version 1.1.0 with this PR.

### DIFF
--- a/specification/OpenMetrics.md
+++ b/specification/OpenMetrics.md
@@ -124,6 +124,8 @@ Exemplars are references to data outside of the MetricSet. A common use case are
 
 Exemplars MUST consist of a LabelSet and a value, and MAY have a timestamp. They MAY each be different from the MetricPoints' LabelSet and timestamp.
 
+In OpenMetrics 1.0.0 Exemplars are only supported for Counter MetricPoints and for Histogram Buckets. OpenMetrics Text format 1.1.0 loosens this restriction: Exemplars are supported on all time series.
+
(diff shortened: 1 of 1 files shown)


jonatan-ivanov (approved):

kakkoyun (approved):

MaxGabriel:
Hi, is the plan to merge this? Saw it was sitting with one approval and was wondering if the consensus changed on it or something

(was working on a client library and figuring out what metrics to support exemplars for)

jonatan-ivanov:
@MaxGabriel I'm not part of the approvers but the Prometheus server already has this capability; if your client library provides exemplars you should see them on the backend, see: https://github.com/prometheus/prometheus/pull/11984
(Not sure about non-Prometheus backends.)

PR #294, opened 2025-01-28
Add Github Actions for Golang/Python workflows

ArthurSens:
I'm basically copying what is configured in Circle CI, but now using Github Actions


bwplotka:
We might need to dive deeper on why... `make test-impl cmd-parser-text=echo` is failing, can check later.

PR #319, opened 2026-05-27
Remove CircleCI

ArthurSens:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,99 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - name: Set up Go
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        with:
+          go-version-file: src/go.mod
```
SuperQ: So, I discovered that this setup-go feature is broken. It installs _exactly_ the version specified by the `go` directive.

ArthurSens: and isn't that what we want? 🤔

ArthurSens:
I'm really struggling to understand the CI failures on these git diffs 😭

.github/workflows/ci.yml
SuperQ: No, if the go.mod says `1.25.0` it will not use `1.26.x`.

ArthurSens:
OKeeee, CI is running as expected now. kdrfc is failing, but seems like it has always been failing 🤔

roidelapluie:
Not all tests have been merged I think

ArthurSens:
> Not all tests have been merged I think

Could you elaborate? I'm not sure I understand what you mean

roidelapluie:
I meant that the old CircleCI job test-success does not seem to be present in the new GitHub Actions workflow. It used to run make test-impl cmd-parser-text=echo.

ArthurSens:
> I meant that the old CircleCI job test-success does not seem to be present in the new GitHub Actions workflow. It used to run make test-impl cmd-parser-text=echo.

Ah, right. Yeah, that's a weird one. Looking at the old CircleCI config, the job definition existed, but it was never scheduled to run 🤔 

I guess one equivalent here would be to define the workflow and make it available only through `workflow_dispatch`. I guess I'll do that 🤷

ArthurSens:
I'm also wondering how useful this workflow actually is (all of them, not only the `make test-impl cmd-parser-text=echo` one)... Since the repo migration from the OpenMetrics org to Prometheus org, CI has been failing consistently. The spec is not even hosted here anymore, but in prometheus/docs.

Maybe we should just get rid of CircleCI and not re-add them to GitHub Actions?

What do OM folks think nowadays? @bwplotka @dashpole @krajorama @ywwg

bwplotka:
We can kill it IMO

ArthurSens:
Alright, the PR just removes CircleCI now. All we need to do is remove the required check.

Can anyone with permissions do it, or teach me how to do it? :)

ArthurSens:
Builds disabled in CircleCI

